PR #13, opened 2018-05-29
Add minified item images

Kaptard:
Some still require proper scaling, although very few from what I've seen.
Mods are currently provided with their raw image, adding mod frames will probably be implemented alongside rivens eventually.


TobiTenno:
👍 cool

Issue #81, opened 2019-05-14
New relics missing drop locations


ibbmarsh:
The relics added for Loki & Volt Prime's unvaulting are missing drop locations. I am not sure when it started, but it is still missing as of v1.290.0 in repo and npm.

See Relic.json, entries for Lith O2, Meso O3, Neo V8, and Axi L4. These are all missing the prop "drops". None of the other unvaulted relics have this issue.

Note that running the build locally caused the drop locations to populate correctly, so it could just be a one-time hiccup.

TobiTenno:
Yeah, it's probably just missing on those builds, and I know Kaptard is working on a refactor, so we'll see about it after that.

Kaptard:
It's fixed in the [refactor/build](https://github.com/WFCD/warframe-items/tree/refactor/build) branch. Will close this issue when merged.

Kaptard:
Merged and will be available on 1.294.0

Issue #22, opened 2018-06-06
Arcanes aren't tracked


Kaptard:
It seems all arcanes except the new ones (including arcane helmets) are missing from Arcanes.json. Not sure which exports file they're in either. This shouldn't be hard to implement - Just find the export URL and add the required condition to the export type in `scraper.addCategory`.

TobiTenno:
By virtue of how they're made, I don't think arcanes will show up as an export, because they don't show anywhere in the mobile app and would, theoretically, otherwise show up in the aranes and relics manifest

TobiTenno:
I think i've got it sorted now for all those i could find in our data... I'll keep looking to see if i can find the helmets

TobiTenno:
Looks like we definitely have Arcane helmets in our skins data, so i think that's sufficient

PR #82, opened 2019-05-15
Refactor build script (Resolves #78 #81)

### LINKED ISSUE #78: Missing images
I just noticed that some images are missing. According to my finding, lots of the missing images are from `Resources.json` (espacailly those plain/vallis related resources).

Kaptard:
Refactor changes:
- Split up data fetching, data processing, data storing methods
- Put a small wrapper around the progress bar so we don't repeat ourselves as much
- Added a warning file that will tell us about data that is definitely missing
- Put all data definitions into /config
- Removed a fuckton of unnecessary complexity
- Improved performance a bit by getting rid of unnecessary loops over all items
- Added a quickFilter method for test builds (this one does the bare minimum so builds would pass, around 90% faster, but lacks data)
- Add components of components (up to one level deep)(currently commented out to limit output size)
- Components now hold a reference to their parent when saved as individual item
- Added a small guide for other developers

Unrelated to the refactor but important:
- Moved the bot to a VPS, so builds should be there much faster than before.


build/build.js
```diff
@@ -1,45 +1,100 @@
...
+      if (chunk.category === 'Recipes') continue // Skip blueprints
+
+      for (let i = 0; i < chunk.data.length; i++) {
+        const item = chunk.data[i]
+
+        if (result[item.category]) {
+          result[item.category].push(item)
         } else {
-          return res
+          result[item.category] = [item]
```
TobiTenno: category is an array now? or am i misreading this?

config/relicGrades.json
```diff
@@ -0,0 +1,6 @@
+[
+  { "id": "Bronze", "tier": "Intact" },
```
TobiTenno: might wanna refactor "tier" to "refinement".

tier also refers to the "age" or Lith/Meso/Neo/Axi quality of relics.

data/json/Quests.json
```diff
@@ -74,7 +74,6 @@
         "name": "Chains of Harrow: Update 21.5.0",
         "date": "2017-08-16T20:44:07Z",
         "url": "https://forums.warframe.com/topic/832153-chains-of-harrow-update-2150/",
-        "imgUrl": "http://n8k6e2y6.ssl.hwcdn.net/uploads/da8dd0d10932c10f50a5d6ed3d8e7cc0.jpg",
```
TobiTenno: did patchlog drop all the image urls?

data/warnings.json
```diff
@@ -0,0 +1,8 @@
...
...YPH", "BAST 50 GLYPH", "BADNEWSBARON GLYPH", "BACKYARDISTV GLYPH", "BABYCHAN TH GLYPH", "BBSCHAINWARDEN GLYPH", "AZRAEL112 GLYPH", "AYWI GLYPH", "AURORASTARR GLYPH", "AUNGELECETTE GLYPH", "ASHI GLYPH", "ARGONSIX GLYPH", "ANNOYING KILLAH GLYPH", "ANGRYICEBERG GLYPH", "ANJETCAT GLYPH", "ALYEKK GLYPH", "ALTHANITHENOBLE GLYPH", "ALEXANDRALIVE GLYPH", "AJINGOM GLYPH", "AEONKNIGHT86 GLYPH", "ADMIRAL BAHROO GLYPH", "ADIKDARKCERO GLYPH", "ADELFOS SELENE GLYPH", "ACCESSIBLEGAMER GLYPH", "A H R GLYPH", "13ANGTV GLYPH", "ILLUMINA JUDICIUM", "WYRMIUS", "FRAME FIGHTER", "HAPPY ZEPHYR", "EXODIA BRAVE", "EXODIA VALOR", "PAX SOAR", "EXODIA MIGHT", "EXODIA HUNT", "PAX SEEKER", "EXODIA TRIUMPH", "EXODIA FORCE", "VIRTUOS TEMPO", "VIRTUOS SHADOW", "VIRTUOS STRIKE", "VIRTUOS FURY", "VIRTUOS NULL", "VIRTUOS GHOST", "VIRTUOS SURGE", "VIRTUOS FORGE", "VIRTUOS SPIKE", "VIRTUOS TROJAN", "MAGUS HUSK", "MAGUS FIREWALL", "MAGUS ELEVATE", "MAGUS NOURISH", "MAGUS REPLENISH", "MAGUS REPAIR", "MAGUS VIGOR", "MAGUS MELT", "MAGUS ACCELERANT", "MAGUS DRIVE", "MAGUS CLOUD", "MAGUS GLITCH", "MAGUS DESTRUCT", "MAGUS ANOMALY", "MAGUS REVERT", "MAGUS OVERLOAD", "MAGUS CADENCE", "MAGUS LOCKDOWN", "PAX CHARGE", "PAX BOLT" ]
```
TobiTenno: did we remove the types for the arcanes? i'm seeing a lot of them in here, as well as primary/secondary weapons

package.json
```diff
@@ -41,7 +41,7 @@
     "node-progress": "^0.1.0",
     "requestretry": "^2.0.2",
     "semantic-release": "^15.13.12",
-    "sharp": "^0.20.8",
+    "sharp": "0.20.5",
```
TobiTenno: is this for the drone issue we hit earlier?

config/polarities.json
```diff
@@ -0,0 +1,22 @@
...
+}, {
+  "id": "AP_PRECEPT",
+  "name": "Penjaga"
+}, {
+  "id": "AP_WARD",
+  "name": "Unairu"
+}, {
+  "id": "AP_UMBRA",
+  "name": "Umbra"
+}]
```
TobiTenno: add....
```json
{
  "id": "AP_UNIVERSAL",
  "name": "???"
}
```

This matches the text shown on those rivens. I think "None" would also make sense. I know there's a new aura polarity as well, but i don't know what we'd find the `AP_XXXXX` name for it under.

data/warnings.json
TobiTenno: I really like the idea of the file overall, though.

data/json/Quests.json
Kaptard: That's the result of the old JSONs. Didn't push the new build yet. Should check if it's actually there then though

data/warnings.json
Kaptard: Arcanes never worked, see #22 
Where are you seeing them though? I can't find them looking for `Arcane`

package.json
Kaptard: Pinned it because 0.20.8 requires image libraries that don't exist on ubuntu (or at least I wasn't able to find them). Specifically it's libwebp.so.7

TobiTenno: ah, interesting

data/warnings.json
TobiTenno: I had a fix for those on a previous build that picked up most of them...

data/json/Quests.json
TobiTenno: I did a clean local build. They seem like they're there for some, including that one, specifically.

data/warnings.json
TobiTenno: Rip, i never got that in or merged. I'll take a look after this is all in.

Kaptard:
Regarding the performance fixes, we'll have to rewrite the addDropRate method as it takes 10 times as much time as the next longest method (addPatchlogs), which takes ~30 times as long as the other methods. Luckily both of these are cached, so I'll skip that overhaul in this PR and look into it at a later point.

Issue #58, opened 2018-12-06
TypeScript typings


will-janz:
I'm making an Angular app using this data (thank you btw) and will be typing out the weapons, pets, frames, etc. I'd like to see types here, I'd be willing to help create them (can't promise any timeline though).

How often does the schema of the data change? Would types be something on the radar and viable?

Thanks!

TobiTenno:
The data schema changes pretty rarely. If you'd like to create typings, feel free to do so and submit a pr.

I don't know that we have explicit plans for them at the moment.

will-janz:
Gotcha. Upon further usage this seems to need more than just types (Angular doesn't like the usual method of importing, though I could be doing it wrong). Closing this, and I'll raise a TS Compatibility issue later on.

Thanks for the info!

PR #83, opened 2019-05-27
Added Typescript type declaration file

Drakmyth:
Built this file for use in my own app and after seeing #58 not resulting in an official file I figured I'd push this back up. From within a Typescript app, the library can be used like so:
```
import Items from 'warframe-items';

const items = new Items({ category: ['All'] });
```
or if you need other types
```
import Items, { ItemsOptions } from 'warframe-items';

const options: ItemsOptions = { category: ['All'] };
const items = new Items(options);
```

I tested the validity of this by installing [typescript-json-schema](https://github.com/YousefED/typescript-json-schema) and [ajv](https://github.com/epoberezkin/ajv) and running
```
typescript-json-schema ./index.d.ts Items --out items.schema.json --noExtraProps --required --strictNullChecks
```
to generate a json-schema and then
```
ajv validate -s ./items.schema.json -d ./data/json/All.json --all-errors --verbose
```
to validate All.json against it, and tweaking the types until no more errors were reported. This is probably a good candidate for an automated test, but I'm not familiar enough with mocha to write this myself just now.

Of note are the string unions making up the bottom half of the file. I created these manually but it should be fairly easy to automate the generation of these so that as new values are added these types remain accurate.

I am a little concerned by some of the data having inconsistent types (for example, "vaulted" can be a `true`/`false` boolean or the string `"N/A"`) but I'm not sure if the scraper has any kind of ETL process to normalize the data or not and didn't want to start touching that if it doesn't.

Finally I should note that while this works for how the library exists today, it would be somewhat nicer if interfaces for each type of item could be split out and the main `Item` type be defined as a union of those interfaces. I started to go down that road, but there is so much data and with all the optional fields and such I suspect a generation process for this file would need to be in place before that is feasible, and this is a good first step.

If there are any clarifications I can provide or questions I can answer, please ask! Happy to contribute to such a great project!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,6 @@
 declare module 'warframe-items' {
     export default class Items extends Array<Item>{
-        constructor(options: ItemsOptions);
+        constructor(options: ItemsOptions, ...items: Item[]);
     }
 
     interface ItemsOptions {
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,461 @@
+declare module 'warframe-items' {
+    export default class Items extends Array<Item>{
+        constructor(options: ItemsOptions);
+    }
+
+    interface ItemsOptions {
+        category: Category[]
+    }
+
+    interface Item {
+        uniqueName: string;
+        name: string;
+        polarity?: Polarity;
+        rarity?: Rarity;
+        baseDrain?: number;
+        fusionLimit?: number;
+        upgradeEntries?: UpgradeEntry[];
+        availableChallenges?: AvailableChallenge[];
+        secondsPerShot?: number;
+        damagePerShot?: number[];
+        magazineSize?: number;
+        reloadTime?: number;
+        totalDamage?: number;
+        damagePerSecond?: number;
+        trigger?: Trigger;
+        description?: string;
+        hexColours?: Colour[];
+        accuracy?: number;
+        criticalChance?: number;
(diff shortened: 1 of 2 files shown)


Drakmyth:
Unrelated, I believe the ci failure prior to my marking this PR as ready for review is due to the ci process not supporting [Draft Pull Requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/). Just wanted to call that out.

TobiTenno:
That's largely the reason it doesn't pick up on CI for the draft, but that's kind of intended, that way it doesn't run a ton extra processes when ultimately drafts are like "not for our eyes" yet.

TobiTenno:
Also, dunno if this changed, but the last time i went about building out the defs for a project, it was SOP to add it to the package.json "types", could you add that for me?

Drakmyth:
> Also, dunno if this changed, but the last time i went about building out the defs for a project, it was SOP to add it to the package.json "types", could you add that for me?

Looks like if the main script is index.js and the type declaration is index.d.ts then adding "types" to package.json is not required, but is still recommended. I'll make that change this evening and get it pushed up.

TobiTenno:
> I'll make that change this evening and get it pushed up.

Sounds good, while it's probably unnecessary for most, it will definitely help. I should probably look at copying a lot of this over to wfcd/api-spec for those applicable bits, too

PR #85, opened 2019-06-20
Patch badge for current update

TobiTenno:
add automated badge update to show the latest update this has built with

(yes, i typo'd the branch name)


TobiTenno:
the badge doesn't look right on that branch.

should i just remove the logo?

Kaptard:
Yeah, the badge doesn't even load for me because of a 414 response (url too long). Do we have to provide the entire image data or do they also accept links?

TobiTenno:
I think it has to be the base64'd data. i'll see if i can find a way to shorten it

TobiTenno:
works, it's just a bad aspect ratio now

Kaptard:
Are you sure we wanna remove the WFCD banner? I doubt people will notice that badge as much as the banner, and I feel like guiding other devs to the organization has rather high priority.

About the order of badges, maybe it makes sense to put the update badge in front or behind the npm badge, as it's certainly more important for the end user than the build state or dependencies.

TobiTenno:
I'm fine with whichever order

Kaptard:
What about the WFCD banner though? Was there any particular reason to use the badge instead?

TobiTenno:
If you think the banner makes it more effective, let's go with that.

PR #87, opened 2019-06-24
Tradable overrides

TobiTenno:
User brought up that the "tradable-ness" of [`argon scope`](https://api.warframestat.us/mods/search/argon%20scope) and [`primed vigor`](https://api.warframestat.us/mods/search/primed%20vigor) were both wrong.

### DIFF
--- a/config/overrides.json
+++ b/config/overrides.json
@@ -115,5 +115,11 @@
   },
   "/Lotus/Upgrades/Mods/Warframe/AvatarMissionSpecificResistanceIce": {
     "isExilus": true
+  },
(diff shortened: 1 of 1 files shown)


TobiTenno:
@Kaptard could we look at making better types for mods? like `warframe mod`, `rifle mod`, `secondary mod`, `shotgun mod`, etc., to make this a bit easier to filter? since most mods, minus the prime ones from daily tribute will be tradeable?

TobiTenno:
it would render this set of changes unnecessary

PR #89, opened 2019-06-25
enemy data parsing

TobiTenno:


index.js
```diff
@@ -1,18 +1,56 @@
...
+      ...defaultOptions,
+      ...options
+    }
 
     // Add items from options to array. Type equals the file name.
     // Tradable determines if we should use sub-folder or stay on root.
     for (let category of this.options.category) {
+      // Ignores the individual enemy categories.
+      // This will need manual addenda if more enemy factions are added.
+      if (this.options.ignoreEnemies && category !== 'Enemy') continue
```
Kaptard: Shouldn't this be `category === 'Enemy'`, so we'd skip that category when enemies are ignored?

TobiTenno: Yeah, it should. I'll fix that

PR #93, opened 2019-07-18
Move to new API (resolves #91)

### LINKED ISSUE #91: Move to new Mobile API
As pointed out in #90 the mobile app no longer uses the old MobileExport API, but instead gets its data from 
```content.warframe.com/PublicExport/Manifest/<Resource_with_id>```

The id can be retrieved from here
```origin.warframe.com/origin/00000000/PublicExport/index_en.txt.lzma```

Thanks @AlyoshaVasilieva, I couldn't figure out how to get the right number for the origin file without pulling the app every time and it didn't even occur to me that all 0's would work lol

### PR DESCRIPTION
**Pull request by Kaptard**

### CODE REVIEW

**Comment by Kaptard** [comment]
There's more deletions than additions (probably only) due to me not updating my warframe-patchlogs version when running the build. But the bot should fix that once it's pushed out. This version won't be published.

**Comment by TobiTenno** [comment]
nah, i can get through it pretty fast.

File: build/scraper.js [hunk 1]
```diff
@@ -123,23 +148,11 @@ class Scraper {
...
   async fetchVaultData () {
     const bar = new Progress('Fetching Vault Data', 1)
-    const vaultData = (await get('http://www.oggtechnologies.com/api/ducatsorplat/v2/MainItemData.json')).data
+    const vaultData = (await getJSON('http://www.oggtechnologies.com/api/ducatsorplat/v2/MainItemData.json')).data
 
     bar.tick()
     return vaultData
   }
-
-  applyOverrides (item) {
```
TobiTenno: any reason we dropped overrides?

Kaptard: It's just duplicate code that I forgot to remove in the last refactor. Overrides are being done in parser.js, they shouldn't be in scraper.js

PR #102, opened 2019-08-01
Recategorized Pets category, resolves #101

### LINKED ISSUE #101: Recategorize some items from Pets.json
So in the past few weeks I was pretty involved in the data this repo produces (great work btw.) and I think there are some inconsistencies in the Pets.json.
I expected only Pets itself from this file, but there are also a bit more unrelated items in there. I created an exclusion list for my project and maybe it makes sense to apply these here also.

`['KubrowPetFood', 'BlankTraitPrint', 'EggHatcher', 'CatbrowGeneticSignature', 'KubrowEgg', 'KubrowPetCollarItem']`
All of these items from the Pets.json are probably better mapped to Misc imo.

Also

- Add Moa's to the file `misc.filter(m => m.uniqueKey.includes('MoaPetHead'))`
- Maybe add Venari too

Xzandro:
This should align more to the Pets.json's purpose

### DIFF
--- a/build/parser.js
+++ b/build/parser.js
@@ -403,13 +403,13 @@ class Parser {
         else if (item.slot === 5) item.category = 'Melee'
         else if (item.slot === 0) item.category = 'Secondary'
         else if (item.slot === 1) item.category = 'Primary'
+        else if (item.uniqueName.includes('/MoaPetHead')) item.category = 'Pets'
         else item.category = 'Misc'
         delete item.isArchwing
         break
 
       case 'Resources':
-        if (item.type === 'Pets') item.category = 'Pets'
-        else if (item.type === 'Specter') item.category = 'Gear'
+        if (item.type === 'Specter') item.category = 'Gear'
         else if (item.type === 'Resource') item.category = 'Resources'
         else if (item.type === 'Fish') item.category = 'Fish'
         else if (item.type === 'Ship Decoration') item.category = 'Skins'
--- a/build/parser.js
+++ b/build/parser.js
@@ -284,7 +284,7 @@ class Parser {
     const types = require('../config/itemTypes.json')
 
     for (let type of types) {
-      if (item.uniqueName.includes(`/${type.id}`)) {
+      if (item.uniqueName.includes(type.id)) {
         item.type = type.name
         break
       }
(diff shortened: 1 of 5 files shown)


build/parser.js
```diff
@@ -403,13 +403,13 @@ class Parser {
         else if (item.slot === 5) item.category = 'Melee'
         else if (item.slot === 0) item.category = 'Secondary'
         else if (item.slot === 1) item.category = 'Primary'
+        else if (item.uniqueName.includes('/MoaPetHead')) item.category = 'Pets'
```
TobiTenno: this should be a reorg in the types parser, preferrably

TobiTenno: instead, add `MoaPetHead` as a key that gets the `Pet` type in [item-types.json](https://github.com/WFCD/warframe-items/blob/development/config/itemTypes.json)

build/parser.js
```diff
@@ -403,13 +403,13 @@ class Parser {
...
         else item.category = 'Misc'
         delete item.isArchwing
         break
 
       case 'Resources':
-        if (item.type === 'Pets') item.category = 'Pets'
```
TobiTenno: please restore the pets bit, that'll be needed to make sure they stay as part of the pets category, we should re-categorize the `Game/KubrowPet` and `Game/CatbrowPet` as types through `item-types.json` so they correctly get picked up that way. You may need to make sure they're above `Pet` to make sure they trigger first, but i think it's cleaner in case other things get added that should be misc or other categories.

I think we should also add better types for the collars and cosmetics that we can find listed, since those, while not pets, are closer to skins

TobiTenno: Revising after discussing in dms:

Remove `KubrowPet` from type names and use `PetPowerSuit` instead

config/itemTypes.json
```diff
@@ -1,136 +1,139 @@
...
-  "id": "Ships",
+  "id": "/Ships",
   "name": "Orbiter"
 }, {
-  "id": "Ship",
+  "id": "/Ship",
   "name": "Extractor"
 }, {
-  "id": "CatbrowPet",
+  "id": "KubrowPetPowerSuit",
```
TobiTenno: does `PetPowerSuit` match both?

Xzandro: Helios also has PetPowerSuit in the unqueName.

TobiTenno: nvm, it also matches something for helios

PR #108, opened 2019-09-04
fix: added Augment type

### LINKED ISSUE #107: Warframe Augment Mods listed as non tradable
A lot of the warframe augment mods (https://warframe.fandom.com/wiki/Warframe_Augment_Mods), if not all (I didn't test everything), are listed as `tradable: false` even though they are tradable.

Examples:

- Guided Effigy
- Mesa's Waltz
- Seeking Shuriken
- Fracturing Crush
- Vampire Leech

Nakroma:
resolves #107

### DIFF
--- a/build/parser.js
+++ b/build/parser.js
@@ -439,7 +439,7 @@ class Parser {
   addTradable (item) {
     const tradableTypes = ['Upgrades', 'Fish', 'Key', 'Focus Lens', 'Relic', 'Rifle Mod',
       'Secondary Mod', 'Shotgun Mod', 'Warframe Mod', 'Companion Mod', 'Archwing Mod', 'K-Drive Mod',
-      'Melee Mod']
+      'Melee Mod', 'Augment']
     const untradableTypes = ['Skin', 'Medallion', 'Extractor', 'Pets', 'Ship Decoration']
     const tradableRegex = /(Prime|Vandal|Wraith|Rakta|Synoid|Sancti|Vaykor|Telos|Secura)/i
     const untradableRegex = /(Glyph|Mandachord|Greater.*Lens|Sugatra)/i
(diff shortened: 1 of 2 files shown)


TobiTenno:
looking at some of the item data, i think this route won't be perfect, because it looks like it also includes things like [Ambush](https://warframe.fandom.com/wiki/Ambush) (ID: `/Lotus/Types/Sentinels/SentinelPrecepts/GhostAugmentCard`), which is just a normal precept.

TobiTenno:
I don't know that anything else will be perfect, cause this definitely looks like the most sure way

TobiTenno:
Potentially might want to make it "Augment Card" just to be in line with the others?

TobiTenno:
I also wonder if it'd be better to add an `isAugment` boolean on the object and check that for the tradeable check, then we could just override the value for ambush to be `false` for `isAugment`

Nakroma:
Ambush is still tradable tho, so it would correctly catch at least that, at least for now it's just a wrong type (which it is anyway currently, it would just change it)

TobiTenno:
very true...
I just don't know that de-typing them as warframe mods will make this an overall good fix.... sorry, just trying to think through and think if i could've written out my initial notes better.

TobiTenno:
hmmmm.... what would you think of doing it the boolean on the object way, and adding better types for sentinel mods so that it also gets marked as tradeable properly?

Nakroma:
sounds like a good idea

TobiTenno:
(@Nakroma hey, sorry for the pingalicious ping... you still down to do that part? I might pick up the branch otherwise and work on it a bit if you're too busy)

Nakroma:
I'm currently back in uni + some other things, so I don't sadly. Feel free to pick it up, if you have the time :)

TobiTenno:
will do. ty for contributing all that you did, it's a giant help.

PR #110, opened 2019-10-07
lexicographic sort for item drop locations

esunder:
Adding lexicographical sort for item drop locations instead of by drop chance
https://github.com/WFCD/warframe-items/issues/105

PR #116, opened 2019-10-09
sorting without data json committed

### LINKED ISSUE #105: Sort by name or Drop Chance? (or both?)
The data seems to thrash a good bit because of the sorting we do on drop chances.

Right now it looks like we're sorting by drop chance, but i'm questioning if we should sort by chance or by something more static like the location so that the order in the array doesn't change so much and the user can sort as needed, but with less changes on our data for otherwise unneeded reasons (specifically when they're received in a different order, but the same drop chance, causing changes to data ordering without any actual data changes)

Alternatively, drop rate, then sub-sort by location? this would provide the value that's present now, but without the thrashing we're getting with lexicographically randomly reordering

esunder:
Fixed lint errors and removed data/json from the commit

closes #110
resolves #105

### DIFF
--- a/build/parser.js
+++ b/build/parser.js
@@ -506,10 +506,17 @@ class Parser {
       if (data.length) item.drops = data
     }
 
-    // Sort by drop rate
     if (item.drops) {
-      item.drops.sort((a, b) => b.chance - a.chance)
+      item.drops.sort(this.comparator)
     }
+  }
+
+  // Compares drop locations for items lexicographically by chance + location + rotation + rarity
+  comparator (dropA, dropB) {
+    // Build a key for comparison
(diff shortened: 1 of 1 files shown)


build/parser.js
```diff
@@ -506,12 +506,19 @@ class Parser {
...
 
-    // Sort by drop rate
     if (item.drops) {
-      item.drops.sort((a, b) => b.chance - a.chance)
+      item.drops.sort(this.comparator)
     }
   }
 
+  // Compares drop locations for items lexicographically by chance + location + rotation + rarity
+  comparator (dropA, dropB) {
```
TobiTenno: you're gonna hate me, but could you call it a `dropComparator` instead, so it doesn't get confused with anything else we may want to compare in the future?

PR #117, opened 2019-10-09
apply overrides for components

### LINKED ISSUE #97: Kavasa Prime Kubrow Collar missing ducats
https://warframe.fandom.com/wiki/Kavasa_Prime_Collar

Band, buckle, and blueprint are 45, 65, 45 ducats according to wiki but not listed in this repo's data.

esunder:
resolves #97
1. Adding overrides for components and applying to the Kubrow Collar components

### DIFF
--- a/build/parser.js
+++ b/build/parser.js
@@ -2,6 +2,12 @@ const Progress = require('./progress.js')
 const previousBuild = require('../data/json/All.json')
 const watson = require('../config/dt_map.json')
 const bpConflicts = require('../config/bpConflicts.json')
+const overridesJson = require('../config/overrides.json')
+const gradesJson = require('../config/relicGrades.json')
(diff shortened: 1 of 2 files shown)


esunder:
done. but curious why are the requires inlined like that? what benefit does it have to do it that way? are they modified while the script is running and need to be reloaded?

TobiTenno:
nope, they aren't reloaded even if they're required more than once, due to them being loaded into memory once, but they're only needed once and it keeps them from being bloat at the top with unnecessarily longer names for what little they're around for

TobiTenno:
as an aside, if you use `resolves` or `closes` instead of `Addresses` in the pr description, it'll close the corresponding issues automatically with the PR instead of needing to be in the commit message 👌

PR #129, opened 2019-12-30
Bugfix: Add missing areaAttack

major0:
areaAttack information missing from various weapons.  This is relevant
to all weapons with a radial damage value such as the Kulstar, Angstrum,
etc.

When looking into this information it was noticed that the damagePerShot
array includes the damage information for the radial Damage, but does
not include relevant information to help distinguish radial damage from
primary damage.  This is relevant as weapons with Radial Damage have
double-gains from all damage mods (adding Corrosion is applied to the
Primary damage and again via Radial Damage). In effect, if a weapon
supports radial damage, then the damagePerShot array is unusable for
calculating DPS of the weapon.

### DIFF
--- a/build/parser.js
+++ b/build/parser.js
@@ -665,6 +665,7 @@ class Parser {
   addWeaponWikiaData (item, wikiaItem) {
     const damageTypes = require('../config/damageTypes.json')
     item.ammo = wikiaItem.ammo
+    item.areaAttack = wikiaItem.areaAttack
     item.channeling = wikiaItem.channeling
     item.chargeTime = wikiaItem.chargeTime
     item.damage = wikiaItem.damage
--- a/index.d.ts
+++ b/index.d.ts
@@ -86,6 +86,7 @@ declare module 'warframe-items' {
         sex?: Sex;
         sprint?: number;
         passiveDescription?: string;
+        areaAttack?: AreaAttack;
         secondary?: Secondary;
         secondaryArea?: SecondaryArea;
         statusChance?: number;
@@ -171,6 +172,19 @@ declare module 'warframe-items' {
         blast?: number;
         corrosive?: number;
         radiation?: number;
+    }
+
+    interface AreaAttack {
+        name: string;
+        status_chance?: number;
+        radius?: number;
+        blast?: number;
+        heat?: number;
+        radiation?: number;
+        damage: string;
+        pellet?: Pellet;
+        duration?: number;
+        speed?: number;
     }
 
     interface SecondaryArea {


TobiTenno:
pretty sure areaAttack is already part of the general damage attacks....

build/parser.js
```diff
@@ -665,6 +665,7 @@ class Parser {
   addWeaponWikiaData (item, wikiaItem) {
     const damageTypes = require('../config/damageTypes.json')
     item.ammo = wikiaItem.ammo
+    item.areaAttack = wikiaItem.areaAttack
```
TobiTenno: you'll also need to add it on the actual object if you're doing it this way

major0:
> pretty sure areaAttack is already part of the general damage attacks....

Curiously, radial damage is not calculated as part of direct damage for weapons such as the Kulstar and the Angstrum.  Damage mods apply damage to the rockets which include all damage calculations.  Upon impact with the target a Radial Attack is calculated (which benefits from damage mods again), and in the case of the Kulstar, pellets (mines) are deployed (3 per rocket), which have a secondaryDamageArea.

I am not aware of any examples of weapons where Radial Damage is applied as part of direct damage calculations.  I suspect Radial Damage is a carry-over from the old Damage 1.0 system in which there was an Explosive Damage type which hit every part of the body independently (and gained benefits for explosive damage against the head).  With Damage 2.0 only the Torso receives the Radial Damage (no more headshot multiplier insanity), but the damage is still applied twice (thus the Kulstar is among the highest DPS weapons in the game).

Again, this information shows up in the game UI, but is not reflected in the mobile UI.

TobiTenno:
Ok, that's fine.
Still need to add it to the object that gets exported.

major0:
> Ok, that's fine.
> Still need to add it to the object that gets exported.

I may need some guidance here.  When I run `npm run build` the data shows up as expected in their respective `data/*.json` files.  I guess these results are not reflective of the data export you are referring to.

index.d.ts
```diff
@@ -86,6 +86,7 @@ declare module 'warframe-items' {
         sex?: Sex;
         sprint?: number;
         passiveDescription?: string;
+	areaAttack?: AreaAttack;
```
TobiTenno: this line looks differently indented.... could you make sure it's consistent?

major0: Yah, looks like 1-space for the unified diff prefix and 8-spaces for the original code, and mine is the `+` prefix plus a hard-tab (dang vim defaults). Guess I need to expand tabs into spaces for this code-base.

TobiTenno: yeah, sorry about that.

TobiTenno:
@major0 i mean where it's applied to a weapon... maybe i'm mixing it up. i'll run it again on my end and check for it. if you want, in a second commit, upload the data, and then i'll let you know when that's been checked, and you can remove that commit from your branch?

TobiTenno:
nvm, tracked down where i wasn't seeing it in parallel with what you were doing. thanks.

index.d.ts
major0: Not a problem, I didn't catch that the code doesn't have hard-tabs.  I will adjust my editor for this repository.

PR #158, opened 2020-10-14
Added missing types to Items

WarNote:
Added missing types.
---

### Considerations
- Does this contain a new dependency?  No
- Does this introduce opinionated data formatting or manual data entry? No
- Does this pr include the updated data files in a separate commit that can be reverted for a clean code-only pr? No
- Have I run the linter through `npm run lint`? Yes
- Is is a bug fix, feature request, or enhancement? Enhancement

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -97,6 +97,45 @@ declare module 'warframe-items' {
         repairRate?: number;
         capacityMultiplier?: number[];
         specialities?: string[];
+        showInInventory?: boolean;
+        systemIndex?: number;
+        systemName?: string;
+        nodeType?: number;
+        missionIndex?: number;
+        factionIndex?: number;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -271,6 +316,21 @@ declare module 'warframe-items' {
         locTag?: string;
     }
 
+    enum ProductCategory {
```
TobiTenno, line 319: not sure that this is correct, I think this would be a type like the others below it... I don't think they're accessible as you're using them here. maybe I don't fully understand typing...

WarNote: It was meant to be an union type but apparently quicktypes generated it wrongly, we can just remove it.

TobiTenno (approved):

PR #170, opened 2021-01-31
Shallow clone

dimon222:
Added shallow clone to reduce clone efforts/speed up clone

Based on:
https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

### Considerations
- Does this pr include the updated data files in a separate commit that can be reverted for a clean code-only pr? **No**
- Have I run the linter through `npm run lint`? **Yes**
- Is is a bug fix, feature request, or enhancement? **Enhancement**

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,7 @@ RUN apk add --update --no-cache git openssh make gcc g++ python curl libpng-dev 
 # Clone repo
 RUN mkdir -p /app/warframe-items \
   && cd app \
-  && git clone https://github.com/WFCD/warframe-items \
+  && git clone --depth 1 --single-branch --branch=development https://github.com/WFCD/warframe-items \
   && cd warframe-items \
   && npm install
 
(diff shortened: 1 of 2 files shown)


TobiTenno:
@dimon222 for the sake of the tests, could you add a data generation run in?

TobiTenno:
I also don't know that this will fix the issue with drone, as i think killian has to manually apply or pull changes into the docker portion... but we can try!

docker/Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apk add --update --no-cache git openssh make gcc g++ python curl libpng-dev
 # Clone repo
 RUN mkdir -p /app/warframe-items \
   && cd app \
-  && git clone https://github.com/WFCD/warframe-items \
```
TobiTenno, line 10: i am worried as this will only ever build development, not master... i guess if the whole process was changed to exclude master, this wouldn't be a problem

TobiTenno:
i'll go ahead and merge this first to see if it'll properly kick off the task in the next few days, otherwise i'll go forward with rescoping package

Issue #171, opened 2021-02-01
No falloff typings for typescript


Tc-001:
**Describe the bug**
There are no typescript typings for Item.AreaAttack.falloff

**To Reproduce**
1. Open visual studio code
2. Create a index.ts file
3. paste in the following:
```ts
import Items from 'warframe-items'
const items = new Items({ category: ['Primary'] })

let this_line_will_error = items[0].areaAttack.falloff
```
**Expected behavior**
There is no typescript error, as falloff is a possible attribute

**Screenshots**
![image](https://user-images.githubusercontent.com/55956895/106498116-99690c00-64c7-11eb-8117-938128fc9cf3.png)
![image](https://user-images.githubusercontent.com/55956895/106499068-c8cc4880-64c8-11eb-93a4-12040d15ea4a.png)

**Versions:**
 - Node.js version: v14.15.4
 - npm version: 6.14.10
 - Version 1.1245.0
- Access Method: npm package

**Additional context**
This should be a simple fix of adding the necessary typings to the index.d.ts file
If needed, I can attempt to fix it and send in a pull request.

TobiTenno:
feel free to submit a PR. I don't run anything ts with this right now, so i have no good way of adding tests for typings. if you know of a good typings test action/framework, i'd be happy to try it.

Tc-001:
Sure, turns out that is is much easier than I thought to make it.

I do not have experience with testing typescript, so I don't know how much I can help.
But one thing that I can think of is to run the raw json file through something similar to http://json2ts.com/ or https://www.npmjs.com/package/json-schema-to-typescript and that can be used to check, or maybe build the .d.ts file depending how smart the package is. (if that is something that you wish to do)

TobiTenno:
the problem is that, due to the data mirroring what DE provides, new fields will always have to be manually added, as i understand typescript and DE's non-passive changes.

PR #172, opened 2021-02-02
Adds falloff typings to index.d.ts

Tc-001:
Fixes issue #171, where no typescript typings were provided for Item.AreaAttack.falloff

closed \#171

---

### Reproduction steps

Open visual studio code
Create a index.ts file
paste in the following:
```ts
import Items from 'warframe-items'
const items = new Items({ category: ['Primary'] })

let this_line_will_error = items[0].areaAttack.falloff
```
---

This adds typings to the index.d.ts file, to describe the attribute Item.AreaAttack.falloff

### Considerations
- Does this introduce opinionated data formatting or manual data entry? **No, don't think so**
- Does this pr include the updated data files in a separate commit that can be reverted for a clean code-only pr? **Not sure**
- Have I run the linter through `npm run lint`? **No**
- Is is a bug fix, feature request, or enhancement? **Enhancement**

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -243,13 +243,15 @@ declare module 'warframe-items' {
         pellet?: Pellet;
         duration?: number;
         speed?: number;
-        falloff?: {
+        falloff?: Falloff;
+    }
+    
+    interface Falloff {
             start: number;
             end: number;
             reduction: number;
-        };
     }
-
+    
     interface SecondaryArea {
         name: string;
         status_chance?: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -245,13 +245,13 @@ declare module 'warframe-items' {
         speed?: number;
         falloff?: Falloff;
     }
-    
+
     interface Falloff {
             start: number;
             end: number;
             reduction: number;
     }
-    
+
     interface SecondaryArea {
         name: string;
         status_chance?: number;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -243,6 +243,11 @@ declare module 'warframe-items' {
         pellet?: Pellet;
         duration?: number;
         speed?: number;
+        falloff?: {
+            start: number;
+            end: number;
+            reduction: number;
+        };
```
TobiTenno, line 250: this should be broken out into another interface, shouldn't it? (at least from the pattern others have used on typings in here)

Tc-001: Oh right, this was how I fixed it in my implementation, but maybe it would be better to break it out

index.d.ts
```diff
@@ -243,13 +243,15 @@ declare module 'warframe-items' {
...
+    }
+    
+    interface Falloff {
             start: number;
             end: number;
             reduction: number;
-        };
     }
-
+
```
TobiTenno, line 254: drop the trailing spaces?

Tc-001: Oh right, definetly
(now if I resolve this will it do it for me...)

TobiTenno: nah, i'd have to add a suggestion

TobiTenno: ah, looks like you force pushed 👍

TobiTenno:
just as a heads up, we're currently having issues with our automated releases, so it might take a while for this to get released

Tc-001:
Sure, good luck on this!

PR #262, opened 2021-10-18
Fix Various TS Typings Issues

### LINKED ISSUE #212: continous-integration type checking script
Sometimes, someone can make a mistake in the typings, and commit something that won't even compile.
Examples: 
- https://github.com/WFCD/warframe-items/commit/3955a934c7e8375efb00470e0c23172ce0ab8420
- A more complex error: https://github.com/WFCD/warframe-items/commit/9836f4b931293e7240d4574db1fe601c5b4e8fa9

To not let these commits pass in pull requests, i suggest a continous-integration script that checks it, the most simple solution would just try to compile the typings in some way, the most advanced would generate typings from the generated data and compare with the current typings. In the first example the simple approach would solve it, but in the second only the advanced approach would be able to detect it.

### LINKED ISSUE #243: [Bug] Arch-Guns have null ammo rather than undifined ammo.
### Describe the bug

The discription of the typings of Item don't correctly match to what's present on Archguns.

### NodeJS Version

v16.7.0

### npm Version

yarn 1.22.11

### warframe-items Version

1.1249.135

### Access Method

npm Package (Default)

### What happened?

Arch-Guns specify ammo as `"ammo": null` rather undefined (absent) which is not the same as the typing listed for `Item`

Drakmyth:
closes #212 
closes #243
along with a variety of other unreported type issues

Also added [typescript-json-schema](https://github.com/YousefED/typescript-json-schema) and [ajv-cli](https://github.com/ajv-validator/ajv-cli) as dev-dependencies and created an npm script command so the typings can be easily validated going forward. 

It looks like typescript-json-schema may be deprecated in favor of [ts-json-schema-generator](https://github.com/vega/ts-json-schema-generator), but I haven't looked into that yet and figured it would be cleaner to make that migration in a separate PR since the tool used here works and was used to create the original typings file and validate all changes thus far.

### DIFF
--- a/.github/workflows/static.yaml
+++ b/.github/workflows/static.yaml
@@ -32,6 +32,7 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - run: npm install
     - run: npm run build
+    - run: npm run typings
   test:
     name: Test
     runs-on: ubuntu-latest
@@ -46,4 +47,3 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - run: npm install
     - run: npm test
-    - run: npm run typings


Drakmyth:
The npm script added here may also be useful to create a CI check to ensure typings are up-to-date as requested by #212.

TobiTenno:
Since you're already in there, on the CI builds, could you kindly add the script in the workflows?

Drakmyth:
> really great overall. just small stuff and one last connecting spot. could you add a step into our [static](https://github.com/WFCD/warframe-items/blob/master/.github/workflows/static.yaml#L35) and [build](https://github.com/WFCD/warframe-items/blob/master/.github/workflows/build.yaml#L25) workflows, to run your type verifications after the build step on each?

I'll definitely do that, but I'd like to do it as a separate PR. Specifically, the way the current generator handles union types is via an older, deprecated syntax that ajv flags doesn't like. Rather than loosening the validation, I'd rather use the updated syntax, but as near as I can tell there isn't a way to tell the current generator to do that. This means when you run the command, you'll see it print:

```
strict mode: use allowUnionTypes to allow union type keyword at "#/properties/flight" (strictTypes)
strict mode: use allowUnionTypes to allow union type keyword at "#/properties/damage" (strictTypes)
strict mode: use allowUnionTypes to allow union type keyword at "#/definitions/RadialDamage/properties/damage" (strictTypes)
```

These errors will prevent the workflow step from being successful. We could work around it by telling it to report these failures as warnings, but since I want to update to the new generator anyway I'd rather just fix it right.

TobiTenno:
this is actually doubly weird because i've pushed with lockfile v2 before and this is the first time i'm seeing it fail on node -14

PR #264, opened 2021-10-20
fix: don't keep element | proc when undefined

### LINKED ISSUE #263: [Bug] Slam Attack element and proc being populated with "undefined" string
### Describe the bug

"element" and "proc" fields in weapon slam attack definition are often set to the string "undefined" rather than actual undefined or null values

### NodeJS Version

N/A

### npm Version

N/A

### warframe-items Version

v1.1250.34

### Access Method

github download

### What happened?

Searching All.json for the string "undefined" reveals 313 occurrences, the vast majority (if not all) of which are in the Slam Attack definition of weapons. Specifically on the "element" and "proc" fields.

TobiTenno:
closed #263 

---

### Reproduction steps
1. Regen data
2. Check Melee data
3. "undefined" proc and elements won't exist anymore

---

You can see my fix in the weapon transformation script where we generate the attack structures

### Considerations
- Does this pr include the updated data files in a separate commit that can be reverted for a clean code-only pr? **[Yes]**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,7 +23,6 @@ jobs:
     - run: npm install
     - run: npm i -S warframe-patchlogs@latest
     - run: npm run build
-    - run: npm run typings
     - uses: stefanzweifel/git-auto-commit-action@v4
       with:
         skip_dirty_check: false


Drakmyth:
Type checking being put to the test right out of the gate 😄 

Need to update the RadialDamage definition on index.d.js lines 265

from 
```
interface RadialDamage {
        damage: number | string;
        element: Element | 'undefined';
        proc: Element | 'undefined';
        radius: number;
}
```

to
```
interface RadialDamage {
        damage: number | string;
        element?: Element
        proc?: Element
        radius: number;
}
```

Though maybe it should be nullable and `Element | string`? Just looking at the transformer logic which seems to allow for string values.

TobiTenno:
i'm wondering, @Drakmyth, if we should turn on pretypings for the normal "builds" so they're fixed when the data is auto-updated and doesn't fail automatic builds

PR #291, opened 2021-12-08
fix: update inconsistent polarities

### LINKED ISSUE #289: 🚀 Inconsistent casing in polarities key between warframes and weapons
No

Change the casing to match between weapons and warframes.

**Additional context**
The problem is in the following polarities: **"umbra", "Umbra", "madurai", "Madurai",  "vazarin", "Vazarin", "naramon" and "Naramon".**

Example:
`{ name: 'Atlas', polarities: [ 'naramon', 'vazarin', 'madurai' ] }`
`{ name: 'Agkuza', polarities: [ 'Vazarin' ] }`

_Note: changed my mind about being a bug, please remove the bug label_

TobiTenno:
closes #289 

---

### Reproduction steps
1. Update & unify polarity maps
2. Force polarities to lower case

---

### Considerations


build/wikia/transformers/polarities.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  Bar: 'naramon',
+  V: 'madurai',
+  D: 'vazarin',
+  U: 'umbra',
+  Ability: 'zenurik',
+  R: 'unairu'
```
AyAyEm, line 7: Missing aura polarity.
```suggestion
  R: 'unairu'
  Aura: 'aura'
```

build/wikia/transformers/transformWarframe.js
```diff
@@ -2,23 +2,22 @@
...
-  R: 'unairu'
-}
+const POLARITIES = require('./polarities')
 
 const transformPolarities = ({ Polarities, AuraPolarity }, targetWeapon) => {
   const outputFrame = { ...targetWeapon }
   if (AuraPolarity) {
-    outputFrame.auraPolarity = POLARITIES[AuraPolarity]
+    outputFrame.auraPolarity = (POLARITIES[AuraPolarity] || AuraPolarity || '').toLowerCase()
+    if (outputFrame.auraPolarity && !outputFrame.auraPolarity.length) outputFrame.auraPolarity = undefined
```
AyAyEm, line 11: Poor design in case there isn't a transformation for a polarity

i think in this case the build should fail, maybe with a more descriptive error than cannot read properties of undefined.
```suggestion
    outputFrame.auraPolarity =POLARITIES[AuraPolarity].toLowerCase()
```

build/wikia/transformers/transformWarframe.js
```diff
@@ -2,23 +2,22 @@
...
+    if (outputFrame.auraPolarity === 'none') outputFrame.auraPolarity = undefined
   }
   if (Polarities) {
-    outputFrame.polarities = Polarities.map(polarity => POLARITIES[polarity])
+    outputFrame.polarities = Polarities.map(polarity => {
+      let out
+      out = (POLARITIES[polarity] || polarity || '').toLowerCase()
+      if (out && !out.length) out = undefined
+      return out
+    }).filter(p => p)
```
AyAyEm, line 20: Same as above

AyAyEm (changes requested):

build/wikia/transformers/transformWarframe.js
TobiTenno: No, actually, this is in the case where we don't have a transform because we don't need one. Fallbacks by design aren't lazy if the source data is already correct. Don't need to map something that just needs a text transformation to be right. 

If it needs a mapping anyway, we don't want that to fail the whole build process, because if it's wrong, someone will notice, and the wiki has already stopped using their aliases, so the mapping is just in case they missed an old one.

build/wikia/transformers/transformWarframe.js
TobiTenno: Same response as above.

build/wikia/transformers/transformWarframe.js
TobiTenno: The requested change would fail on anything with an undefined mapping, which would fail builds.

AyAyEm (approved):

Issue #276, opened 2021-11-22
:rocket: Weapon Component Resource Costs


bmgolley:
The costs to build weapon components, like the Ambassador parts, are missing. There are resource costs to build the Necramech components as well as Necraloid's arch-guns, but all other weapon components have no separate entry.

TobiTenno:
There's an itemCount property on components. not sure what you're missing.

https://api.warframestat.us/weapons/search/ambassador?only=name,components
```json
[
  {
    "components": [
      {
        "uniqueName": "/Lotus/Types/Recipes/Weapons/WeaponParts/CrpArSniperBarrel",
        "name": "Barrel",
        "description": "A weapon-crafting component.",
        "itemCount": 1,
        "imageName": "barrel.png",
        "tradable": false,
        "drops": [
          {
            "chance": 0.05,
            "location": "Aurax Atloc Raknoid",
            "rarity": "Common",
            "type": "Ambassador Barrel Blueprint"
          }
        ]
      },
      {
        "uniqueName": "/Lotus/Types/Recipes/Weapons/CrpArSniperBlueprint",
        "name": "Blueprint",
        "description": "Bring negotiations to an instant conclusion with this Corpus assault rifle. Switch effortlessly between charged explosive sniper shots and rapid-fire electrical assault mode.",
        "itemCount": 1,
        "imageName": "blueprint.png",
        "tradable": false,
        "drops": [
          {
            "chance": 0.0833,
            "location": "Neptune/Enkidu Ice Drifts (Skirmish), Rotation C",
            "rarity": "Rare",
            "type": "Ambassador Blueprint"
          },
          {
            "chance": 0.0833,
            "location": "Pluto/Fenton's Field (Skirmish), Rotation C",
            "rarity": "Rare",
            "type": "Ambassador Blueprint"
          },
          {
            "chance": 0.0833,
            "location": "Venus/Luckless Expanse (Skirmish), Rotation C",
            "rarity": "Rare",
            "type": "Ambassador Blueprint"
          },
          {
            "chance": 0.1667,
            "location": "Veil/Lu-Yan (Skirmish), Rotation C",
            "rarity": "Uncommon",
            "type": "Ambassador Blueprint"
          }
        ]
      },
      {
        "uniqueName": "/Lotus/Types/Recipes/Weapons/WeaponParts/CrpArSniperReceiver",
        "name": "Receiver",
        "description": "A weapon-crafting component.",
        "itemCount": 1,
[...]

PR #293, opened 2021-12-09
perf: trim package size

TobiTenno:
start changes for #276 

---

### Reproduction steps
1. remove All.json from package (saves ~33MB)
2. Generate array of all categories on init (minus `All` and `i18n`)
2. Remove `['All']` as the default for category and replace with an array of all categories
3. Dedupe category array so we don't dupe items in "this"
4. Minify all but All.json
5. Persist build data across ci jobs

---

You can see my changes in `index.js` and test in `test/index.js`

### Considerations
- Is is a bug fix, feature request, or enhancement? **[Feature/Maintenance]**

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -105,7 +105,7 @@ class Build {
 
     // All.json (all items in one file)
     all.sort(sort)
-    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all))
+    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all, 2))
     fs.writeFileSync(path.join(__dirname, '../data/json/i18n.json'), stringify(i18n))
 
     return all
--- a/build/build.js
+++ b/build/build.js
@@ -100,12 +100,12 @@ class Build {
     for (const category in categories) {
       const data = categories[category].sort(sort)
       all = all.concat(data)
-      fs.writeFileSync(path.join(__dirname, `../data/json/${category}.json`), JSON.stringify(JSON.parse(stringify(data))))
+      fs.writeFileSync(path.join(__dirname, `../data/json/${category}.json`), stringify(data))
     }
 
     // All.json (all items in one file)
     all.sort(sort)
-    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all))
+    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all, 2))
     fs.writeFileSync(path.join(__dirname, '../data/json/i18n.json'), stringify(i18n))
 
     return all
--- a/build/build.js
+++ b/build/build.js
@@ -100,12 +100,12 @@ class Build {
     for (const category in categories) {
       const data = categories[category].sort(sort)
       all = all.concat(data)
-      fs.writeFileSync(path.join(__dirname, `../data/json/${category}.json`), stringify(data))
+      fs.writeFileSync(path.join(__dirname, `../data/json/${category}.json`), JSON.stringify(JSON.parse(stringify(data))))
     }
 
     // All.json (all items in one file)
     all.sort(sort)
-    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all, 2))
+    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all))
     fs.writeFileSync(path.join(__dirname, '../data/json/i18n.json'), stringify(i18n))
 
     return all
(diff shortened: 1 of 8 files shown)


test/index.js
```diff
@@ -1,46 +1,63 @@
...
+
 beforeEach(() => {
   Items = require('../index')
 })
 afterEach(() => {
   decache('../index.js')
 })
+// after(() => {
+//   console.error(`Average init time: ${(inits.reduce((a, b) => a + b, 0) / inits.length).toFixed(0)}`)
+// })
```
TobiTenno, line 25: commented to allow debugging more later... cause test code, i don't care if it stays in, personally

index.js
```diff
@@ -32,14 +32,25 @@
...
+
 let i18n = {}
 try {
   i18n = require('./data/json/i18n.json')
 } catch (ignored) {
   // can only happen in really weird stuff, and we're already defaulting, so it's ok
 }
-const defaultOptions = { category: ['All'], i18n: false, i18nOnObject: false }
+
+const raw = fs.readdirSync(path.join(__dirname, './data/json/'))
```
AyAyEm, line 46: raw doesn't mean anything here,  i would suggest `categoryFiles`

index.js
```diff
@@ -32,14 +32,25 @@
...
+const ignored = ['All', 'i18n']
+const all = raw
```
AyAyEm, line 48: all also doesn't mean much, i would suggest `categories`, `allCategories` or `possibleCategories`

AyAyEm (changes requested):

AyAyEm:
I think this pr is more of a refactor than feature, also we're not changing the size of all.json we're removing it from the package, i think something like `trim all.json from package` or `ignore all.json in package` would fit better.

index.js
TobiTenno: sure

index.js
TobiTenno: sure

TobiTenno:
yeah, i actually did that just by allowing specification of spaces

build/build.js
```diff
@@ -105,7 +105,7 @@ class Build {
 
     // All.json (all items in one file)
     all.sort(sort)
-    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all))
+    fs.writeFileSync(path.join(__dirname, '../data/json/All.json'), stringify(all, 2))
```
AyAyEm, line 108: Why would All.json be the single exception, not being minified?

TobiTenno: because it's the only json:
- not in the build
- used to validate types

TobiTenno: And validating types is what needs, seemingly, an unminified `All.json` to work correctly. I'm not 100% sure why.

build/build.js
```diff
@@ -100,7 +100,7 @@ class Build {
     for (const category in categories) {
       const data = categories[category].sort(sort)
       all = all.concat(data)
-      fs.writeFileSync(path.join(__dirname, `../data/json/${category}.json`), stringify(data))
+      fs.writeFileSync(path.join(__dirname, `../data/json/${category}.json`), JSON.stringify(JSON.parse(stringify(data))))
```
AyAyEm, line 103: Could you add a comment explaining why it needs to double parse it?

build/stringify.js
```diff
@@ -9,7 +9,7 @@ const isPrimitive = obj => obj === null || ['string', 'number', 'boolean'].inclu
 const isArrayOfPrimitive = obj => Array.isArray(obj) && obj.every(isPrimitive)
 const format = arr => `^^^[ ${arr.map(val => JSON.stringify(val)).join(', ')} ]`
 const replacer = (key, value) => isArrayOfPrimitive(value) ? format(value) : value // objectsort(value)
-const expand = str => str.replace(/"\^\^\^(\[ .* \])"/g, (match, a) => a.replace(/\\"/g, '"')).replace(/\\\\"/g, "'")
-const stringify = obj => expand(JSON.stringify(Array.isArray(obj) ? dedupe(obj) : obj, replacer, 2))
+const expand = str => str.replace(/"\^\^\^(\[ .* ])"/g, (match, a) => a.replace(/\\"/g, '"')).replace(/\\\\"/g, "'")
+const stringify = (obj) => expand(JSON.stringify(Array.isArray(obj) ? dedupe(obj) : obj, replacer, 2))
```
AyAyEm, line 13: I think it's better to remove the parentheses to keep the project consistent
```suggestion
const stringify = obj => expand(JSON.stringify(Array.isArray(obj) ? dedupe(obj) : obj, replacer, 2))
```

AyAyEm (changes requested):

AyAyEm:
Also would be possible to edit the description for this pr? Just for keeping things clean.

build/stringify.js
TobiTenno: I'm planning a styling update pass before I do the ES upgrade. Can I do it then?

TobiTenno:
> Also could you edit the description for this pr? Just for keeping things clean.


To include or remove what, specifically?

AyAyEm:
> To include or remove what, specifically?

We also minfied the json data, changed from fix to feat to perf and refactored two action scripts. But just adding that we minfied the data would be fine.

build/stringify.js
AyAyEm: Sure if you plan to change it anyway it won't be a problem.

PR #304, opened 2022-01-10
Replace colors with chalk

Bitti09:
Replace colors with chalk  to avoid  things like   https://github.com/Marak/colors.js/issues/285 in the future 

---
### Reproduction steps
 None

---

If you know you know
https://github.com/Marak/colors.js/issues/285
### Considerations


SlayerOrnstein:
Should we consider doing this for other projects as well?

TobiTenno:
> Should we consider doing this for other projects as well?

I've got a pending concern out on the status one, i'm gonna say i'd rather not just switch for the sake of switching, as colors has worked find apart from this hiccup, and is under new maintenance now

Issue #381, opened 2022-10-27
:rocket: Hash data from DE to determine if updates are needed


TobiTenno:
**Is your enhancement request related to a problem? Please describe.**
Essentially, we publish too many releases

**Describe the upgrade you'd like**
Update data if DE's data has updated, so we need to hash their origin data. I'd think the hash in their URL is adequate, possibly, but we'd need to store a hash per-manifest file, but that shouldn't be that bad.

example, in the url...
`http://content.warframe.com/PublicExport/Manifest/ExportWeapons_en.json!00_fJwSHnUxX1w6f4exHJ0Z9g`
we need to store a key of `ExportWeapons_en.json` with a key of `00_fJwSHnUxX1w6f4exHJ0Z9g`, so we'd just have to check if `00_fJwSHnUxX1w6f4exHJ0Z9g` has changed when we pull new copies of `ExportWeapons_en.json` if we want to update it. If no changes are detected, don't run any of the build process.

Ideally, we'd be able to just update the parts that are updated with changed hashes, but our build process wouldn't allow that.

**Describe alternatives you've considered**
Hashing all of our data after build, but this seems pretty pointless, as at that point we're including wiki data that changes violently and uncontrollably with urls and such.

**Additional context**
We just want less builds

TobiTenno:
parallel consideration:
given that we're making a mechanism to _not_ build, we need a mechanism to bypass this check to _force_ local builds.
so generally you'd want an environment variable, so we'd need to add documentation to the README to specify how to bypass the build hash checking

PR #382, opened 2022-10-27
feat(build): hash checking DE data before build

AyAyEm:

closed \#381

- [x] Basic hash checking to stop the build process
- [x] Force flag to the build script: `-f` `--force` which ignores the hash checking
- [x] Support for environment variable to force the build
- [x] Update readme

---

### Reproduction steps

---

### Considerations

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const _ = require('lodash');
 const fs = require('fs/promises');
 const path = require('path');
 const crypto = require('crypto');
@@ -14,6 +15,7 @@ const scraper = require('./scraper');
 const parser = require('./parser');
 const imageCache = require('../data/cache/.images.json');
 const exportCache = require('../data/cache/.export.json');
+const locales = require('../config/locales.json');
 
 const allowedCustomCategories = ['SentinelWeapons'];
 
@@ -22,8 +24,16 @@ const get = async (url, binary = true) => {
   return binary ? res.buffer() : res.text();
 };
 
+const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg));
+
 class Build {
   async init() {
+    const updatedExportCache = await this.generateExportCache();
+    if (!force && _.isEqual(exportCache, updatedExportCache)) {
+      console.log('Finished, data already up-to-date');
+      return;
+    }
+
     const resources = await scraper.fetchResources();
     /** @type {RawItemData} */
     const raw = {
@@ -50,7 +60,28 @@ class Build {
     for (const warning of Object.keys(parsed.warnings)) {
       warningNum += parsed.warnings[warning].length;
     }
+
+    await fs.writeFile(
+      path.join(__dirname, '../data/cache/.export.json'),
+      JSON.stringify(updatedExportCache, undefined, 1)
+    );
+
     console.log(`\nFinished with ${warningNum} warnings.`);
+  }
+
+  async generateExportCache() {
+    const keyWhitelist = ['Manifest', 'DropChances', 'Patchlogs'];
+
+    const endpoints = await Promise.all([...locales, 'en'].map((locale) => scraper.fetchEndpoints(false, locale)));
+
+    return Object.fromEntries([
+      ...keyWhitelist.filter((key) => key in exportCache).map((key) => [key, exportCache[key]]),
+      ...endpoints
+        .flat()
+        .map((endpoint) => endpoint.split('!00_'))
+        .filter(([key, hash]) => key && hash)
+        .map(([key, hash]) => [key, { hash }]),
+    ]);
   }
 
   /**
--- a/build/build.js
+++ b/build/build.js
@@ -23,7 +23,7 @@ const get = async (url, binary = true) => {
   return binary ? res.buffer() : res.text();
 };
 
-const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg)) || process.env.FORCE !== 'true';
+const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg)) || process.env.FORCE === 'true';
 
 class Build {
   async init() {
--- a/build/build.js
+++ b/build/build.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const fs = require('fs/promises');
 const path = require('path');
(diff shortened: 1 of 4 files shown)


build/build.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const _ = require('lodash');
```
TobiTenno, line 3: let's avoid lodash

AyAyEm: Done, but why we don't like lodash?

build/build.js
```diff
@@ -22,8 +23,16 @@ const get = async (url, binary = true) => {
   return binary ? res.buffer() : res.text();
 };
 
+const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg));
```
AyAyEm, line 26: Is it ok to pass args to build like this?

build/build.js
```diff
@@ -22,8 +23,16 @@ const get = async (url, binary = true) => {
...
+
 class Build {
   async init() {
+    await dataHash.updateExportCache();
+    if (!force && process.env.FORCE !== 'true' && dataHash.isUptodate()) {
+      console.log('Finished, data already up-to-date');
```
AyAyEm, line 32: Maybe a better message here?

build/dataHash.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
```
AyAyEm, line 1: The other file names are just a single word, and this one have two, is this a problem?

build/dataHash.js
```diff
@@ -0,0 +1,57 @@
...
+
+  async updateExportCache() {
+    const endpoints = await this.fetchEndpoints();
+    endpoints
+      .flat()
+      .map((endpoint) => endpoint.split('!00_'))
+      .filter(([key, hash]) => key && hash)
+      .forEach(([key, hash]) => {
+        this.exportCache[key] = { hash };
+      });
```
AyAyEm, line 45: Is anything wrong with long chained methods like this?

build/dataHash.js
```diff
@@ -0,0 +1,57 @@
...
+
+const fs = require('fs/promises');
+const path = require('path');
+const scraper = require('./scraper');
+const exportCache = require('../data/cache/.export.json');
+const locales = require('../config/locales.json');
+
+const exportKeyWhitelist = ['Manifest', 'DropChances', 'Patchlogs'];
+
+class DataHash {
```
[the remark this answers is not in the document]
AyAyEm: Also is it okay to abstract the function context with a class? Like in parser?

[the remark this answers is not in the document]
AyAyEm: Maybe UpstreamHash, DEDataHash, DEHash, ApiHash

build/build.js
```diff
@@ -28,8 +27,8 @@ const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg
 
 class Build {
   async init() {
-    const updatedExportCache = await this.generateExportCache();
-    if (!force && _.isEqual(exportCache, updatedExportCache)) {
+    await dataHash.updateExportCache();
+    if (!force && process.env.FORCE !== 'true' && dataHash.isUptodate()) {
```
TobiTenno, line 31: maybe include the process.env.FORCE check as part of force's initialization

build/dataHash.js
TobiTenno: that seems ok

build/dataHash.js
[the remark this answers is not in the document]
TobiTenno:
```suggestion
class HashManager {
```

build/dataHash.js
```diff
@@ -0,0 +1,57 @@
...
+  constructor() {
+    this.exportCache = Object.fromEntries(
+      exportKeyWhitelist.filter((key) => key in exportCache).map((key) => [key, exportCache[key]])
+    );
+  }
+
+  isUptodate() {
```
TobiTenno, line 18: make sure to comment these, just to mention which caches they do and don't change/update/etc

build/dataHash.js
TobiTenno: i don't have issues with them, so long as they're readable like this

build/dataHash.js
```diff
@@ -0,0 +1,57 @@
...
+      });
+  }
+
+  async fetchEndpoints() {
+    const localeEndpoitns = await Promise.all(
+      [...locales, 'en'].map((locale) => scraper.fetchEndpoints(false, locale))
+    );
+
+    return localeEndpoitns.flat();
+  }
```
TobiTenno, line 54: not sure that this belongs here, since this is more of what we do in the scraper, so if we need to do a static export from there instead, that might make more sense.

TobiTenno (changes requested):

build/build.js
TobiTenno: mostly because i don't want node code to end up like browser-side code and the tendency to pull in mega-libraries like lodash or jquery

build/build.js
TobiTenno: this is fine by me, but we should include the env check in the same line, so
```suggestion
const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg)) || process.env.FORCE === 'true';
```

that is probably too long, so might need to lint locally

build/build.js
```diff
@@ -22,8 +23,16 @@ const get = async (url, binary = true) => {
   return binary ? res.buffer() : res.text();
 };
 
+const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg)) || process.env.FORCE !== 'true';
```
TobiTenno, line 26:
```suggestion
const force = process.argv.slice(2).some((arg) => ['--force', '-f'].includes(arg)) || process.env.FORCE === 'true';
```

TobiTenno (changes requested):
just a smol logical fix

build/build.js
AyAyEm: opsie

AyAyEm: eslint won't let me do this way unfortunately

TobiTenno (approved):

TobiTenno:
want me to rebase, or do you wanna rebase manually?

AyAyEm:
> want me to rebase, or do you wanna rebase manually?

go for it

PR #385, opened 2022-10-29
feat: parse mod sets, avionics, focus upgrades

### LINKED ISSUE #178: Is there any data about components or armaments of RailJack?
I'm a wiki editor from Fandom. If there are relevant data about components or armaments of RailJack, it will be easier to tabulate data according to json files instead of checking one by one in the game. And it is also good for some database websites.

If there are relevant data about components or armaments of RailJack, you could add some information about where they are. You could write them down in README.md.

If there are not, but some repositories in github or other websites have, you could add some links.

**Additional context**
Thanks for listening.

TobiTenno:
closes #178

---

### Reproduction steps
1. Check for the category, push more data from the scraped json
2. Run builds (which should trigger releases) on changes from build files

---

the only weird thing is mods sets are typed as "Mod Set Mods", not that i like it... but i couldn't find a better way yet

### Considerations
- Is is a bug fix, feature request, or enhancement? **[Enhancement]**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,6 +2,12 @@ name: Build
 on:
   schedule:
     - cron: "55 */6 * * *"
+  push:
+    branches:
+      - master
+    paths:
+      - 'build/*.js'
+      - 'config/*.json'
 
 jobs:
   build:
--- a/build/scraper.js
+++ b/build/scraper.js
@@ -79,6 +79,14 @@ class Scraper {
       const raw = await getJSON(`https://content.warframe.com/PublicExport/Manifest/${endpoint}`);
       const data = raw ? raw[`Export${category}`] : undefined;
       bar.tick();
+
+      if (category === 'Upgrades') {
+        const modSets = raw.ExportModSet.map((ms) => ({
+          ...ms,
+          type: 'Mod Set',
+        }));
+        data.push(...modSets, ...raw.ExportAvionics, ...raw.ExportFocusUpgrades);
+      }
       return { category, data };
     };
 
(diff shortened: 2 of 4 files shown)


build/scraper.js
```diff
@@ -79,6 +79,14 @@ class Scraper {
       const raw = await getJSON(`https://content.warframe.com/PublicExport/Manifest/${endpoint}`);
       const data = raw ? raw[`Export${category}`] : undefined;
       bar.tick();
+
+      if (category === 'Upgrades') {
+        const modSets = raw.ExportModSet.map((ms) => ({
+          ...ms,
```
AyAyEm, line 85: It's bad practice to use shortened names when there's no need to, even if it's clear what they should mean.
```suggestion
        const modSets = raw.ExportModSet.map((modSet) => ({
          ...modSet,
```
kudos to you, for not modifying the original modSet object.

.github/workflows/build.yaml
```diff
@@ -2,6 +2,12 @@ name: Build
 on:
   schedule:
     - cron: "55 */6 * * *"
+  push:
+    branches:
+      - master
+    paths:
+      - 'build/*.js'
+      - 'config/*.json'
```
AyAyEm, line 10: Remember to add to the merge commit and/or the description that you changed CI

AyAyEm (changes requested):

.github/workflows/build.yaml
TobiTenno: i can do that! gimme a sec

AyAyEm (approved):

PR #390, opened 2022-11-13
fix: move catchall down

TobiTenno:
addresses...
image

---

### Reproduction steps
1. moved catchall back to the bottom
2. added tests to enforce some melee. i don't have a good way of forcing all of them, cause that'd take a lot of pre-enumeration.

### Considerations
- Does this pr include the updated data files in a separate commit that can be reverted for a clean code-only pr? **[Yes]**


AyAyEm:
Maybe i can work with the typings to enforce type and category, like having more interfaces for specific type category keys. That would prevent this kind of issue and some others.

AyAyEm:
Also i just noticed we need to modify the build workflow and force the build when it's not a cronjob, because like with this modification it won't re-build because the upstream data didn't change.

AyAyEm (approved):
If you wish to also include the changes to CI in this PR, i ask to request the review again.

TobiTenno:
> Maybe i can work with the typings to enforce type and category, like having more interfaces for specific type category keys. That would prevent this kind of issue and some others.

Agreed, but weapons are dicey cause it'd require splitting attacks, I think.

PR #395, opened 2022-11-26
fix: force build where is needed

AyAyEm:
Not building, when build script changes, because of non schedule builds not being forced

---

### Reproduction steps
1. Force build when it's not triggered by schedule
2. Increase build schedule frequency, _because now there shouldn't be a problem with it_

---

### Considerations
- Does this pr include the updated data files in a separate commit that can be reverted for a clean code-only pr? **[no]**

_It should work_

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,7 +1,7 @@
 name: Build
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


TobiTenno (commented):
Looks good

TobiTenno (approved):

PR #396, opened 2022-11-26
feat: build refactor to esm

### LINKED ISSUE #268: :clock1030: Convert warframe-items to ES

AyAyEm:
closes #268

---

### Reproduction steps
1. Refactored the code to ESM without changing any logic _when possible_

---

### Considerations
- Does this contain a new dependency? **[Yes]**


build/parser.mjs
```diff
@@ -1080,4 +1083,6 @@ class Parser {
   }
 }
 
-module.exports = new Parser();
+export const parser = new Parser();
+
+export default parser;
```
TobiTenno, line 1088: i don't really like doing both

build/sortObject.mjs
```diff
@@ -1,16 +1,16 @@
...
-module.exports = (obj) => {
+export function sortObject(obj) {
   return Object.keys(obj)
     .sort()
     .reduce(function (result, key) {
       result[key] = obj[key];
       return result;
     }, {});
-};
+}
```
TobiTenno, line 14: why switch from closure to function?
just name the single closure as a const and export it

build/wikia/transformers/transformPolarity.mjs
```diff
@@ -1,6 +1,4 @@
-'use strict';
-
-const POLARITIES = require('./polarities');
+import { POLARITIES } from './polarities.mjs';
```
TobiTenno, line 1:
```suggestion
import POLARITIES from './polarities.mjs';
```

TobiTenno: use default where possible, not the destructured ones

build/wikia/transformers/transformWarframe.mjs
```diff
@@ -1,8 +1,7 @@
-'use strict';
+import getColors from 'get-image-colors';
+import { image as _image } from 'image-downloader';
```
TobiTenno, line 2: :-1: don't use dangling underscores out front
```suggestion
import { image as getImage } from 'image-downloader';
```

build/wikia/transformers/transformWarframe.mjs
```diff
@@ -28,7 +27,7 @@ const mapColors = async (oldFrame, imageUrl) => {
  * @param {Record<string, Object>} blueprints blueprint objects
  * @returns {Promise<WikiaWarframe>}
  */
-module.exports = async (oldFrame, imageUrls, blueprints) => {
+export async function transformWarframe(oldFrame, imageUrls, blueprints) {
```
TobiTenno, line 30: as elsewhere, don't use functions, use named closures, especially since we're exporting named anyway

TobiTenno (changes requested):

build/.eslintrc.yaml
```diff
@@ -1,3 +1,11 @@
+extends: '@wfcd/eslint-config/esm'
+parserOptions:
+  babelOptions:
+    plugins:
+      - '@babel/plugin-syntax-import-assertions'
```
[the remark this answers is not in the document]
AyAyEm: yeah that's why i already sent a pr upstream

build/parser.mjs
AyAyEm: me too but i default is priority in the eslint config and i don't really like that, i don't like how the default import works, it opens up the opportunity to have two different words meaning the same thing in the code base or one word meaning two things, that's in module scope.

build/sortObject.mjs
AyAyEm: sure i just think function is a more concise way to declare a function when there's no need for the closured feature

build/wikia/transformers/transformPolarity.mjs
AyAyEm: why?

build/wikia/transformers/transformWarframe.mjs
AyAyEm: it seems it was the vscode that did that i didn't realize

build/wikia/transformers/transformPolarity.mjs
AyAyEm: as stated before with destructuring it's more clear that i'm not renaming anything and it's what is expected to be. Different from default export which could two different names in two modules. I think this is bad style and design.

build/parser.mjs
TobiTenno: the word default or the name parser? cause if you just export the default, then you can't do a destructure

build/sortObject.mjs
TobiTenno: yeah, but our styling, i'm pretty sure, specifies avoiding functions entirely

build/.eslintrc.yaml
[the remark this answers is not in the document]
TobiTenno: i'm pretty sure this requires node 18 (for lts), as well, this is something we'd discussed on wfcd/warframe-worldstate-parser#354

build/wikia/transformers/transformPolarity.mjs
TobiTenno: yes, but our styling that i've had in place for a while specifies that we should prefer defaults for exports, so why should we do the opposite for imports?

build/.eslintrc.yaml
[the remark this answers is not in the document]
AyAyEm: Yes it was after but i always test myself and it worked, but it may not  work depending on the version of node 16

build/parser.mjs
AyAyEm: [the export default ](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/prefer-default-export.md)
in the parser it would be okay to remove the default but it wouldn't be in other modules (files)

build/wikia/transformers/transformPolarity.mjs
AyAyEm: > yes, but our styling that i've had in place for a while specifies that we should prefer defaults for exports, so why should we do the opposite for imports?

Maybe i'll open a PR on the styling then.

build/.eslintrc.yaml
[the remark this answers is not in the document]
AyAyEm: https://github.com/nodejs/node/pull/41736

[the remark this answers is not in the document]
TobiTenno: yeah, that's why i'd written the index.mjs how it's written for importing json in the old compatible way.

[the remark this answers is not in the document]
AyAyEm: that's another way to load json but for the future the json loading will be with assertions, i'm just future proofing and since we use at least v16 there shouldn't be any problem with it.

build/build.mjs
```diff
@@ -1,20 +1,18 @@
...
+import minifyJpeg from 'imagemin-jpegtran';
+import fetch from 'node-fetch';
+import sharp from 'sharp';
+
+import Progress from './progress.mjs';
+import stringify from './stringify.mjs';
+import scraper from './scraper.mjs';
+import parser from './parser.mjs';
+import hashManager from './hashManager.mjs';
+import imagesCache from '../data/cache/.images.json' assert { type: 'json' };
```
TobiTenno, line 14: why'd this get renamed?

build/wikia/transformers/elements.mjs
```diff
@@ -13,3 +11,6 @@ module.exports = {
   Gas: 'gas',
   Void: 'void',
 };
+Object.freeze(ELEMENTS);
+
+export default ELEMENTS;
```
TobiTenno, line 16: it's never modified, only used internally. no point doing extra ops every time for something that's already treated in code as a const object with const values...

(same thing on all the other enum mappers)

build/build.mjs
TobiTenno: looks like several got similar treatments. i'd rather not have name refactors that are hard to track on top of the language switchover

TobiTenno (changes requested):

build/.eslintrc.yaml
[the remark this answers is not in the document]
TobiTenno: the future "good way" that was supposed to only be going into v18 of lts and was in unstable until a few feature versions on LTS ago... i don't like using experiments when  they're still in experimental on 16

image

as i mentioned in WFCD/warframe-worldstate-parser#354, i'm fine using them when they're in stable state on those LTS versions, but not until then

build/build.mjs
AyAyEm: it was an opsie that i made, _i blame vscode for that_

AyAyEm:
I made everything again by _hand_ so no renaming this time.

build/.eslintrc.yaml
```diff
@@ -1,3 +1,4 @@
+extends: '@wfcd/eslint-config/esm'
 rules:
   no-console: off
   import/no-extraneous-dependencies: off
```
TobiTenno, line 4: :thinking: was this added because of the other parts? this should be covered by our rule for extraneous deps where we allow dev dependencies to be used in non-spec code :thinking: i'm thinking we need to update the matching in the esm to match .mjs and normal js

TobiTenno: i think v1.3.1 fixes this from being needed

TobiTenno: (i swear this is the last thing.... i think)

TobiTenno: (and rebasing the branch)

AyAyEm: If we're working with esm in the build why the rules should be for javascript?

It's not because of dev dependencies which was also an issue but in the tests, this is for the rule that i changed specifically for esm, esm modules work best when specifying the file extension peer node and browser specs. That's why every local import has the extension of the file generally `.mjs`, but eslint upstream rule wouldn't allow it so i changed from:
```js
'import/extensions': 0
```
to
```js
'import/extensions': ['error', 'always']
```

[the node specifications for esm extension resolution](https://nodejs.org/docs/latest-v16.x/api/esm.html#mandatory-file-extensions)

AyAyEm: I also don't understand why a rebase should be needed

AyAyEm: > 🤔 why was this added? this should be covered by our rule for extraneous deps where we allow dev dependencies to be used in non-spec code 🤔 i'm thinking we need to update the matching in the esm to match .mjs and normal js

or maybe you mean `import/no-extraneous-dependencies: off`? that it should be removed?

TobiTenno: correct, should be able to remove `import/no-extraneous-dependencies: off` with the latest update. i think what you're doing in always giving the extension is fine and makes perfect sense. i was commenting specifically on the line about `import/no-extraneous-dependencies: off`

AyAyEm: makes sense my bad

AyAyEm: The rule isn't exactly right in there it should be `build/**/*.mjs` not exactly sure why. And also shouldn't we put this rules in the eslint config of warframe-items as it's specific to it? Or is there other projects that also benefit from it?

AyAyEm: like
```yaml
  import/no-extraneous-dependencies:
    - error
    - devDependencies:
        - "build/**/*.mjs"
```

TobiTenno: it's related to patchlogs as well, since they both use the same build output folder...

which part doesn't work? is it getting overridden in the esm config and i'm just silly?

TobiTenno: it's odd that it's an issue with it now but didn't have an issue with it before 🤔 let's still update the base config, that's why i have it there :D

TobiTenno: should publish here in a sec

TobiTenno (approved):
Woot

PR #397, opened 2022-11-27
fix(tradable): proof fragment tradable property

AyAyEm:
Set the `Proof Fragment` item tradable property to true, as i t should be

---

### Reproduction steps
1. Add an override in `overrides.json` to set the tradable property to true

---

### Considerations


TobiTenno (approved):
seems like a fairly strightforward one, since there's nothing else we can really go on for tradability here...

Wolfizen:
Thanks :)

PR #412, opened 2023-01-23
fix: set `marketCost` to undefined when string is empty

SlayerOrnstein:
### What did you fix? 
Looks like the wikia doesn't have an in-game market price for all items so instead of an empty string it should be set to undefined.

---

### Reproduction steps

1.  Run build
2. Check Twin Krohkur
3. `marketCost: ""`

---

image

### Considerations

### DIFF
--- a/build/parser.mjs
+++ b/build/parser.mjs
@@ -869,7 +869,7 @@ class Parser {
   addWeaponWikiaData(item, wikiaItem) {
     item.attacks = wikiaItem.attacks;
     item.ammo = wikiaItem.ammo;
-    item.marketCost = wikiaItem.marketCost || undefined;
+    item.marketCost = typeof wikiaItem.marketCost === 'number' ? wikiaItem.marketCost : undefined;
     item.bpCost = wikiaItem.bpCost;
     item.masteryReq = item.masteryReq || wikiaItem.mr;
     item.polarities = wikiaItem.polarities;
--- a/build/parser.mjs
+++ b/build/parser.mjs
@@ -1,10 +1,10 @@
 import _ from 'lodash';
 
 import dedupe from './dedupe.mjs';
+import hashManager from './hashManager.mjs';
 import Progress from './progress.mjs';
 import readJson from './readJson.mjs';
 import tradable from './tradable.mjs';
-import hashManager from './hashManager.mjs';
 
 const previousBuild = await readJson(new URL('../data/json/All.json', import.meta.url));
 const watson = await readJson(new URL('../config/dt_map.json', import.meta.url));
@@ -869,7 +869,7 @@ class Parser {
   addWeaponWikiaData(item, wikiaItem) {
     item.attacks = wikiaItem.attacks;
     item.ammo = wikiaItem.ammo;
-    item.marketCost = wikiaItem.marketCost;
+    item.marketCost = wikiaItem.marketCost || undefined;
     item.bpCost = wikiaItem.bpCost;
     item.masteryReq = item.masteryReq || wikiaItem.mr;
     item.polarities = wikiaItem.polarities;
--- a/test/index.spec.mjs
+++ b/test/index.spec.mjs
@@ -143,65 +143,59 @@ for (const base of ['index.js', 'index.mjs']) {
         assert.strictEqual(matches.length, 1, 'There can be only One');
       });
     });
-  });
-  describe(`${base} integrity`, async function () {
-    this.timeout(10000);
-    after(() => {
-      delete require.cache[base];
-    });
-    it('weapons should only have 1 result for Mausolon', () => {
-      const matches = data.weapons
-        .filter((i) => i.name === 'Mausolon')
-        .map((i) => {
-          delete i.patchlogs;
-          return i;
-        });
-      const dd = dedupe(matches);
-      assert.strictEqual(dd.length, matches.length, 'Before and after dedupe should match');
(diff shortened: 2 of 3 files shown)


SlayerOrnstein:
Also I think vs code is sorting the imports on save, I have no idea how to turn that off

TobiTenno:
Could you add a test for checking that nothing has an empty string as the marketCost?

SlayerOrnstein:
> Could you add a test for checking that nothing has an empty string as the marketCost?

Sure can do

AyAyEm:
Also could you modify the types to `marketCost?: number` in the component and Buildable interfaces?

TobiTenno:
probably need to make sure it's a number if that's the provided type, as well....

AyAyEm:
The test you suggested already does that

TobiTenno:
no, that ensures it's not an empty string or a falsy value, not that it is a number...

AyAyEm:
well the other typing is `marketCost?: number | ''` so if it's not an empty string it should be a number or undefined.

TobiTenno:
Good point.

SlayerOrnstein:
@AyAyEm types have been added.
@TobiTenno I have no idea if I'm doing this testing thing for js right.

test/index.spec.mjs
```diff
@@ -197,5 +197,11 @@ for (const base of ['index.js', 'index.mjs']) {
         assert(warframe?.components?.length > 0);
       });
     });
+
+    it('marketCost should be a number when present', () => {
+      data.items.filter(i => !namedExclusions.includes(i.marketCost)).forEach((item) => {
+        assert(item.marketCost)
+      })
+    })
```
AyAyEm, line 205: I would probably do something more like this, it doesn't make sense to check for excalibur prime in this case also we want a more clear error message for when something goes wrong

```suggestion
    data.items.forEach((item) => {
      it(`${item.name} marketCost should be a number if present`, () => {
        assert(['number', 'undefined'].includes(typeof item.marketCost));
      });
```
or
```suggestion
    data.items
      .filter((item) => typeof item !== 'undefined')
      .forEach((item) => {
        it(`${item.name} marketCost should be a number if present`, () => {
          assert(typeof item.marketCost === 'number');
        });
      });
```

build/parser.mjs
```diff
@@ -869,7 +869,7 @@ class Parser {
   addWeaponWikiaData(item, wikiaItem) {
     item.attacks = wikiaItem.attacks;
     item.ammo = wikiaItem.ammo;
-    item.marketCost = wikiaItem.marketCost;
+    item.marketCost = wikiaItem.marketCost || undefined;
```
AyAyEm, line 872: Could you add a comment here just so we know why exactly we're doing this weird or with undefined? Or make it more clear the intention? Like:
```suggestion
    item.marketCost = wikiaItem.marketCost === '' ? undefined : wikiaItem.marketCost;
```

AyAyEm (changes requested):

build/parser.mjs
TobiTenno: lol, i actually asked him to do it this way. this way if it's ever falsy, we specify exactly the falsy value it should be

AyAyEm: future proof it sure

test/index.spec.mjs
TobiTenno: i would move this into the existing foreach and just unwrap it

TobiTenno: i don't think these tests are working at all anyway, so i'm gonna work on this for you, @SlayerOrnstein, and we'll pair on it later

SlayerOrnstein: Oh okay so I wasn't going crazy because the test above that should have outputted the names for the items being tested and I didn't think anything was limiting the logger.

SlayerOrnstein: > I would probably do something more like this, it doesn't make sense to check for excalibur prime in this case also we want a more clear error message for when something goes wrong

So the reason for this was more to just write something so @TobiTenno can take a look, was never actually supposed to run, sorry about that

TobiTenno (approved):
love it

TobiTenno:
probably easiest to rebase from your end

AyAyEm (approved):

build/parser.mjs
AyAyEm, line 872: I just remembered if marketCost is 0 it'll be assigned to undefined
```suggestion
    item.marketCost = typeof wikiaItem.marketCost === 'number' ? wikiaItem.marketCost : undefined;
```

AyAyEm (changes requested):

build/parser.mjs
TobiTenno: If its 0 it already is undefined for all purposes. If its 0 plat, there's no point knowing its 0...

AyAyEm: some items may be free, i don't remember exactly if it already happened but it may be a thing

AyAyEm (approved):

build/parser.mjs
SlayerOrnstein: Items that are free are always set to 1 credit. I wanted to say that this morning but I was waking up and figured being information complete is just as important.

But to be honest the information for market cost comes from the wikia so you also have to consider the one person who sets the market cost to 0.

TobiTenno (approved):

PR #424, opened 2023-04-11
Fix: Correctly categorize sigils

SnowyLeopard:
### What did you fix? 
Sigils where being marked as skins instead of being put in their own category.
Tracked down the issue being the order of the entry within the `itemTypes.json` file.

Since sigils have unique names like `/Lotus/Upgrades/Skins/Sigils/xxx` it matched `/Skins` first. Now the type is correctly set.

---

### Reproduction steps

1. Run build
2. See that sigils formerly being put in `Skins.json` are now in `Sigils.json` and `Sigils.json` is updated again.
3. For example `Nightwave Sigil` is no longer in `Skins.json`
---

Before:
image

After:
image

### Considerations

### DIFF
--- a/config/itemTypes.json
+++ b/config/itemTypes.json
@@ -17,6 +17,9 @@
   "id": "/Archwing/Melee",
   "name": "Arch-Melee"
 }, {
+  "id": "/Sigils",
+  "name": "Sigil"
+}, {
   "id": "/Skins",
   "name": "Skin"
 }, {
@@ -73,9 +76,6 @@
 }, {
   "id": "/Keys",
   "name": "Key"
-}, {
-  "id": "/Sigils",
-  "name": "Sigil"
 }, {
   "id": "/Game/QuartersWallpapers",
   "name": "Ship Decoration"


config/itemTypes.json
```diff
@@ -16,6 +16,9 @@
 }, {
   "id": "/Archwing/Melee",
   "name": "Arch-Melee"
+}, {
+  "id": "/Sigils",
+  "name": "Sigil"
```
TobiTenno, line 21: this would set the type, not the category

TobiTenno: i'm down with the change, but it's not going to categorize, just change the type (which would probably keep it in Skins.json)

TobiTenno: lol, here's how dumb i am. this would totally get the job done.

https://github.com/WFCD/warframe-items/blob/ce433f807af546fa518042bf4fb20a34f8acf6b9/build/parser.mjs#L567-L569 (thanks to this)

TobiTenno (approved):

PR #425, opened 2023-04-22
feat: Masterable attribute

SnowyLeopard:
### What did you fix? 
Adds a new attribute to items `masterable` denoting whether the item is masterable or not.
Certain categories contain masterable items, these are listed in a new config file `masterableCategories`.
When an item matches the category it will be listed as masterable.

---

### Reproduction steps

---

### Considerations


TobiTenno:
woot, ty, bud! not trying to be a pain, it's just harder to review with that in :D

build/parser.mjs
```diff
@@ -188,6 +189,7 @@ class Parser {
     this.addResistanceData(result, category);
     this.addRelics(result, data.relics);
     this.applyOverrides(result);
+    this.applyMasterable(result);
```
TobiTenno, line 192: this should come before `this.applyOverrides` so that the overrides apply last

index.d.ts
```diff
@@ -52,6 +52,7 @@ declare module 'warframe-items' {
         description?: string;
         releaseDate?: string;
         excludeFromCodex?: boolean;
+        masterable?: boolean;
```
TobiTenno, line 55: given that it's applying on all items, shouldn't be optional, like it should fail types if one of the items is missing this altogether...

```suggestion
        masterable: boolean;
```

test/index.spec.mjs
```diff
@@ -2,6 +2,9 @@ import assert from 'node:assert';
 import { resolve } from 'node:path';
 import { createRequire } from 'module';
 import dedupe from '../build/dedupe.mjs';
+import readJson from '../build/readJson.mjs';
+
+const masterableCategories = await readJson(new URL('../config/masterableCategories.json', import.meta.url));
```
TobiTenno, line 7: we already have a bypass as part of the main load for doing that, should just use the require we already have, as that will read the json... (i.e. don't need `readJson`)

TobiTenno (changes requested):
i like the change overall, just some nit-picks

SnowyLeopard:
> woot, ty, bud! not trying to be a pain, it's just harder to review with that in :D

Haha no problem! I completely understand ;)
I've processed the feedback you had, thanks for that! Makes the code better 😄

PR #443, opened 2023-06-22
fix(images): fetch icon images for abilities

SlayerOrnstein:
### What did you fix? 
Fixes images for abilities not being fetched, also introduces `uniqueName` key to abilities.

Since I'm not super familiar with JS I'm not entirely sure if this approach was the best way to handle it so pointers are welcomed.

---

### Reproduction steps

---

### Considerations

### DIFF
--- a/build/build.mjs
+++ b/build/build.mjs
@@ -190,6 +190,13 @@ class Build {
           await this.saveImage(component, true, duplicates, manifest);
         }
       }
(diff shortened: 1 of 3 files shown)


TobiTenno (approved):

SlayerOrnstein:
Yeah no idea what's going on with the nodes though. Appears before and after the fix

PR #459, opened 2023-08-09
fix(types): explicit export path in package.json

BenjaminPalko:
### What did you fix? 
image
Fixed types not being exported properly through package.json for ESM projects

---

### Reproduction steps
Install warframe-items as a dependency on an ESM Typescript project (Webstorm)

---

### Considerations

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "exports": {
     "types": "./index.d.ts",
     "import": "./index.mjs",
-    "require": "./index.js"
+    "default": "./index.js"
   },
   "scripts": {
     "build": "node ./build/build.mjs",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "exports": {
     "types": "./index.d.ts",
     "import": "./index.mjs",
-    "default": "./index.js"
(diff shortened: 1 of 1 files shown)


TobiTenno:
Your description is misleading. This only applies to ESM projects.

package.json
```diff
@@ -5,8 +5,9 @@
   "main": "index.js",
   "types": "index.d.ts",
   "exports": {
+    "types": "./index.d.ts",
     "import": "./index.mjs",
-    "require": "./index.js"
```
TobiTenno, line 9: Why? No thanks. Please don't remove an existing.

TobiTenno (changes requested):

BenjaminPalko:
> Your description is misleading. This only applies to ESM projects.

Ty, didn't realize it was only ESM.

TobiTenno (approved):

TobiTenno:
uhhhh, looks like some weirdness with types... doubt it's specifically from you, but if you wanna take a look at that error and put a ticket in with the root cause and fix that, can't merge this until that's fixed (sorry)

Issue #482, opened 2023-10-12
:rocket: Inconsistencies in category schema


mdpung:
## **\*\*Please note this is just my observation and feedback based on my experience with using this data set, and meant for food for thought.**\*\*

**Is your enhancement request related to a problem? Please describe.**
I noticed inconsistencies in the way weapons are classified in their respective custom categories.
Let's compare three weapons for example:
A weapon like `Soma` has 
```
{
    "category": "Primary",
    "productCategory": "LongGuns"
}
```

A weapon like `Deconstructor` has
```
{
    "category": "Primary",
    "productCategory": "SentinelWeapons"
}
```

A weapon like `Corvas` has
```
{
    "category": "Primary",
    "productCategory": "Arch-Gun"
}
```
Yet the three belong to the respective categories:
```
Soma - Primary
Deconstructor - SentinelWeapons
Corvas - Arch-Gun
```
When I realized that the `Deconstructor` was not in the `Primary.json` it took me a while to realize this discrepancy.

Another discrepancy I noticed is that Archwing and Tenno have their separate categories for their types of weapons. Ex:

Tenno: `Primary.json` and `Melee.json`
Archwing: `Arch-Gun` and `Arch-Melee`
Sentinel: `SentinelWeapons`

**Describe the upgrade you'd like**
I can understand the more specific sub categories for the robotics and the archwings, so I would like to suggest a few options.

1. Don't change anything and just document in the README.md these differences.
**OR**
2. Change the name of `Primary.json` to something not so broad so it doesn't umbrella the `SentinelWeapon` and `Arch-Gun` exceptions, such as `TennoPrimary`. In addition, you could have `Primaries.json` encompass all the weapons that actually have the category as `Primary`.

As for the second discrepancy, may I suggest:
1. Break `SentinelWeapons` into `SentinelPrimary` and `SentinelMelee`

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

TobiTenno:
Primaries aren't melee. Why would Deconstructor be in primaries?

TobiTenno:
SentinelWeapons is driven off the DE-provided field of ProductCategory to make them easy to find. I'm not gonna break them up further because of your preferences

mdpung:
> Primaries aren't melee. Why would Deconstructor be in primaries?

That's what your data model currently classifies it as. The Deconstructor currently has the category of "Primary" in the dataset, so that should be a question for the devs.

mdpung:
[...]

PR #485, opened 2023-10-17
fix(types): SentinelWeapons typing

mdpung:
### What did you fix? 
SentinelWeapons was incorrectly typed as it was not typed as a Category and should be as it behaves similarly to "Primary" and "Secondary" categories.

---

### Reproduction steps
```
import Items, { Category } from "warframe-items";

// Originally, this instantiation will not work because "SentinelWeapons" is not typed as a Category but should be.
const weaponTypes: Category[] = ["Primary", "Secondary", "Melee", "Arch-Gun", "Arch-Melee", "SentinelWeapons"];
const items = new Items({ category:  });
```

---

image

### Considerations

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -155,7 +155,7 @@ declare module 'warframe-items' {
     }
     interface SentinelWeapon extends Omit<Melee, 'category'> {
         sentinel: true;
-        productCategory: 'Melee';
+        productCategory: 'SentinelWeapons';
         category: 'SentinelWeapons';
     }
     interface ExaltedWeapon extends Omit<Gun, 'category'>, Omit<Melee, 'category'>, Omit<Weapon, 'category'> {
@@ -608,6 +608,7 @@ declare module 'warframe-items' {
         'OperatorAmps' |
         'Pistols' |
         'Sentinels' |
+        'SentinelWeapons' |
         'SpaceGuns' |
         'SpaceMelee' |
         'SpaceSuits' |
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,7 +5,7 @@ declare module 'warframe-items' {
         i18n: BundleofI18nBundle<Locale>;
     }
     interface ItemsOptions {
-        category: Array<Category | 'SentinelWeapons'>;
+        category: Array<Category>;
         ignoreEnemies?: boolean;
         i18n?: boolean|string[];
         i18nOnObject?: boolean;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -607,7 +607,6 @@ declare module 'warframe-items' {
         'Melee' |
         'OperatorAmps' |
         'Pistols' |
-        'SentinelWeapons' |
```
mdpung, line 610: Should be a Category not a ProductCategory

mdpung:
Also a fix for inconsistency mentioned in #482

index.d.ts
TobiTenno: it is a ProductCategory and a Category.

TobiTenno (changes requested):
`SentinelWeapons` is a `ProductCategory`

mdpung:
> * `SentinelWeapons` is a `ProductCategory`
> 
> * `SentinelWeapons` is a valid entry to the `categories` array, but is a `ProductCategory`

I see that the productCategory is defined by DE, so I'll leave it as a productCategory. But since it's also a category, it's not properly typed as a category. I will make the update and see if that's more acceptable.

index.d.ts
mdpung: I will make changes to reflect that properly.

TobiTenno:
that's fine if you do that on your branch, but i'm not going to merge it.

mdpung:
Can I ask why? There's literally bugs in the code.

mdpung:
All the change is, is changing the SentinelWeapon type category from "Primaries" to "SentinelWeapon" (you said it was wrong yourself) and making SentinelWeapon a proper type for category.

TobiTenno:
Because I said it shouldn't be changed. Yes, I'm annoying, sorry. My point is you're crossing streams of a ProductCategory into Category, and I'm telling you I don't want it.

TobiTenno:
This change will break more then it fixes so at the time being this isn't a critical fix

mdpung:
SentinelWeapons was made a Category and ProductCategory by design so I would say those streams were already crossed lol. I'll just do my own data repository, this project seems to be causing me more headaches than what it's worth.

TobiTenno:
it's literally not a category. it's an option for the category argument because there's a file created for it. that's not the same as a member of the Category type

TobiTenno:
your changes weren't even passing the type checks, so i wouldn't have merged it even if i had started agreeing with you that this should get fix immediately

Issue #496, opened 2023-11-25
Every relic is vaulted


Attula15:
According to the json files every relic that the json contains is vaulted.

TobiTenno:
we just inherit vaulted status. if you'd like to replace how we do that, feel free.

FelipiMatheuz:
It's been more than two weeks and still have no solution.
Where this info is collected?

TobiTenno:
https://github.com/search?q=repo%3AWFCD%2Fwarframe-items+%22vaulted+%3D%22&type=code

SlayerOrnstein:
@FelipiMatheuz are you look at a specific relic?

TobiTenno:
@Attula15 please take a look now and see if it's working better for you

Attula15:
@TobiTenno Planning to, but currently I am not near any kind of PC sry.

TobiTenno:
> Planning to, but currently I am not near any kind of PC sry.

@Attula15 any luck?

Attula15:
Yep, its fine. Thank you very much. Sorry for the delay, I have a lot work to do nowdays.

TobiTenno:
Nope, I've got it.

PR #504, opened 2024-01-05
fix: missing vault status for some primes

SlayerOrnstein:
### What did you fix? 
This fixes `vault` missing for some prime warframes. i.e. Nidus Prime was missing `vault` due to data not being available on ogg. Vault status is now retrieved from the wikia and fallsback to ogg.

---

### Reproduction steps

---

image

### Considerations
- Does this introduce opinionated data formatting or manual data entry? **No?**

### DIFF
--- a/build/parser.mjs
+++ b/build/parser.mjs
@@ -193,7 +193,7 @@ class Parser {
     this.addPatchlogs(result, data.patchlogs);
     this.addAdditionalWikiaData(result, category, data.wikia);
     this.addIsPrime(result);
-    this.addVaultData(result, data.vaultData);
+    this.addVaultData(result, data.vaultData, category, data.wikia);
     this.addResistanceData(result, category);
     this.addRelics(result, data.relics);
     this.applyMasterable(result);
@@ -220,7 +220,7 @@ class Parser {
     this.addImageName(result, data.manifest, previous);
     this.addCategory(result, category);
 
-    this.addVaultData(result, data.vaultData);
+    this.addVaultData(result, data.vaultData, category, data.wikia);
 
     return result;
   }
@@ -954,12 +954,21 @@ class Parser {
    * data from "Ducats or Plat".
    * @param {Item} item data to append vault data to
    * @param {VaultData} vaultData to look up data for the {@param item}
+   * @param {module:warframe-items.Category} category of the data
+   * @param {WikiaData} wikiaData from wikia to apply
    */
-  addVaultData(item, vaultData) {
+  addVaultData(item, vaultData, category, wikiaData) {
+    let vaultCategory = category;
+    if (item.type === 'Archwing') vaultCategory = 'archwings';
+
     if (!item.name.endsWith('Prime')) return;
+    if (!['weapons', 'warframes', 'archwings', 'sentinels'].includes(vaultCategory.toLowerCase())) return;
+
+    if (vaultCategory === 'Sentinels') vaultCategory = 'companions';
+    const wikiaItem = wikiaData[vaultCategory.toLowerCase()].filter((i) => i).find((i) => i.name === item.name);
     const target = vaultData.find((i) => i.Name.toLowerCase() === item.name.toLowerCase());
 
-    if (!target) {
+    if (!target && !wikiaItem) {
       const isManuallyExcluded = primeExcludeRegex.test(item.name);
       const isSkin = item.category === 'Skins';
       const isSentinelWeapon =
@@ -971,18 +980,18 @@ class Parser {
       return;
     }
 
-    if (target.ReleaseDate) {
+    if (target?.ReleaseDate) {
       item.releaseDate = this.normalizeOggDate(target.ReleaseDate);
     }
-    if (target.VaultedDate) {
+    if (target?.VaultedDate) {
       item.vaultDate = this.normalizeOggDate(target.VaultedDate);
     }
-    if (target.EstimatedVaultedDate) {
+    if (target?.EstimatedVaultedDate) {
       item.estimatedVaultDate = this.normalizeOggDate(target.EstimatedVaultedDate);
     }
-    if (target.Vaulted) {
-      item.vaulted = target.Vaulted;
-    }
+
+    item.vaulted = target?.Vaulted ?? wikiaItem.vaulted;
+    if (item.vaulted === 'N/A') item.vaulted = true;
   }
 
   /**


build/parser.mjs
```diff
@@ -941,12 +941,20 @@ class Parser {
...
    */
-  addVaultData(item, vaultData) {
+  addVaultData(item, category, wikiaData, vaultData) {
+    if (!['weapons', 'warframes', 'mods', 'upgrades'].includes(category.toLowerCase())) return;
     if (!item.name.endsWith('Prime')) return;
+
     const target = vaultData.find((i) => i.Name.toLowerCase() === item.name.toLowerCase());
 
+    if (item.type === 'Archwing') category = 'archwings';
+    if (category === 'Upgrades') category = 'mods';
```
AyAyEm, line 955: reassignment of function parameters is not very good by design, in this case nothing bad should happen, but could you give another name to category in this context like vaultCategory? And also if you're changing function parameters to use in the rest of the function do it at the first chance you have, it'll be easier to read.

For example
```js
let vaultCategory = category;
if (item.type === 'Archwing') vaultCategory = 'archwings';
if (vaultCategory === 'Upgrades') vaultCategory = 'mods';

if (!['weapons', 'warframes', 'mods', 'upgrades'].includes(vaultCategory.toLowerCase())) return;
if (!item.name.endsWith('Prime')) return;

const target = vaultData.find((i) => i.Name.toLowerCase() === item.name.toLowerCase());
```

build/parser.mjs
```diff
@@ -193,7 +193,7 @@ class Parser {
     this.addPatchlogs(result, data.patchlogs);
     this.addAdditionalWikiaData(result, category, data.wikia);
     this.detectPrime(result);
-    this.addVaultData(result, data.vaultData);
+    this.addVaultData(result, category, data.wikia, data.vaultData);
```
AyAyEm, line 196: you forgot to update the line 223 with the new arguments

build/scraper.mjs
```diff
@@ -1,17 +1,18 @@
...
 import ModScraper from './wikia/scrapers/ModScraper.mjs';
-import WeaponScraper from './wikia/scrapers/WeaponScraper.mjs';
-import WarframeScraper from './wikia/scrapers/WarframeScraper.mjs';
 import VersionScraper from './wikia/scrapers/VersionScraper.mjs';
-import readJson from './readJson.mjs';
+import WarframeScraper from './wikia/scrapers/WarframeScraper.mjs';
+import WeaponScraper from './wikia/scrapers/WeaponScraper.mjs';
 
 import { get, getJSON, retryAttempts } from './network.mjs';
+import ArchwingScraper from './wikia/scrapers/ArchwingScaper.mjs';
```
AyAyEm, line 15: could you revert these changes? It seems that your formatter thought it would be better to rearrenge them 😂

AyAyEm (changes requested):
I only reviewed the code, i will do the review of the data if @TobiTenno approve the idea of adding more wikia dependency

build/parser.mjs
```diff
@@ -968,9 +976,8 @@ class Parser {
     if (target.EstimatedVaultedDate) {
       item.estimatedVaultDate = this.normalizeOggDate(target.EstimatedVaultedDate);
     }
-    if (target.Vaulted) {
-      item.vaulted = target.Vaulted;
-    }
+
+    item.vaulted = wikiaItem.vaulted || target.vaulted;
```
AyAyEm, line 980: is the target.vaulted that unreliable so we need to depend directly from the wikia if available?

build/parser.mjs
SlayerOrnstein: This is actually not allowed in dart either so not sure why I did it here 🤣

build/parser.mjs
SlayerOrnstein: oh

build/parser.mjs
SlayerOrnstein: if you want we could use ogg as the primary and the wikia second

AyAyEm: nah i'm ok having the wikia first

TobiTenno: i think i was saying it in a weird way that wasn't helpful. if ogg is defined, use ogg. if wiki is defined and ogg is not, use wiki... imo.

the fallback language, i think, was making my meaning unclear

build/scraper.mjs
```diff
@@ -12,6 +12,7 @@ import VersionScraper from './wikia/scrapers/VersionScraper.mjs';
 import readJson from './readJson.mjs';
 
 import { get, getJSON, retryAttempts } from './network.mjs';
+import ArchwingScraper from './wikia/scrapers/ArchwingScaper.mjs';
```
TobiTenno, line 15: this should be above ModScraper... not sure why this got imported lie this

SlayerOrnstein: auto import ftw right

build/scraper.mjs
```diff
@@ -134,9 +135,7 @@ class Scraper {
    * @property {string} textureLocation location of the image for the corresponding item
    */
   /**
-   * @typedef {Object} ImageManifest
-   * @property {ImageManifestItem[]} Manifest
-   */
+   * */
```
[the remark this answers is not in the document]
SlayerOrnstein: I'm not even sure what happened here. It looks like it was changed by the formatter but it removed that whole comment... I'm seeing why now and it says because it wasn't declared so I'm deferring to you if you want me to clean up the empty comment or add it back

build/scraper.mjs
```diff
@@ -227,8 +229,7 @@ class Scraper {
 
   /**
    * Formatted date string. Format: "YYYY MM DD"
-   * @typedef {string} OggDateStamp
-   */
+   *    */
```
TobiTenno, line 232: wth is vs code doing. please stop.

TobiTenno (changes requested):
seems fine overall, just some weird formatting on vs code

build/scraper.mjs
SlayerOrnstein: it's actually off. Looks like this was after I ran lint:fix

build/parser.mjs
SlayerOrnstein: so no more

build/scraper.mjs
TobiTenno: right, but lint doesn't normally fix what you've already got in the comment... i'd actually rather we not get rid of our typedef for OggDateStamp...

build/scraper.mjs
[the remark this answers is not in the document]
TobiTenno: I'd prefer it stick around. typedefs exist to create type definitions that are for the internal mechanism of the library in this case for a bunch of stuff that exist for their own sake. i'm not sure why vs code is telling you it's not defined... this is how typedefs work in jsdoc, lol

build/scraper.mjs
```diff
@@ -133,10 +134,6 @@ class Scraper {
    *    identifier corresponding to a {@link module:warframe-items.Item.uniqueName|uniqueName}
    * @property {string} textureLocation location of the image for the corresponding item
    */
-  /**
-   * @typedef {Object} ImageManifest
-   * @property {ImageManifestItem[]} Manifest
-   */
```
TobiTenno, line 139: this is an internal type def and should not be removed

build/scraper.mjs
```diff
@@ -227,7 +227,6 @@ class Scraper {
 
   /**
    * Formatted date string. Format: "YYYY MM DD"
-   * @typedef {string} OggDateStamp
```
TobiTenno, line 230: this is an internal typedef and should not be removed

SlayerOrnstein: Just pushed out the commit that adds it back

build/parser.mjs
```diff
@@ -941,11 +941,20 @@ class Parser {
    * Adds releaseDate, vaultDate and estimatedVaultDate to all primes using
    * data from "Ducats or Plat".
    * @param {Item} item data to append vault data to
+   * @param {module:warframe-items.Category} category of the data
+   * @param {WikiaData} wikiaData from wikia to apply
    * @param {VaultData} vaultData to look up data for the {@param item}
    */
-  addVaultData(item, vaultData) {
+  addVaultData(item, category, wikiaData, vaultData) {
```
TobiTenno, line 948: i'm not sure why the category param is needed, we should already have the category defined from the item

also, generally you add new params at the end of the signature, not in the middle, unless you're adding required parameters before optional parameters.... but all of these are required

SlayerOrnstein: It's not but because I don't understand javascript as much a fair bit was done from what I understood in `addAdditionalWikiaData`, I did want to originally use `item.category` or `item.type` but because I didn't see it being used I assumed there was a reason for it and just added it here, which is about the same explanation for the order actually

TobiTenno: ok, what about the stuff that can be vaulted? you can't vault resources...

SlayerOrnstein: skill issues on my part then. but yeah I got you

TobiTenno: nono, this is definitely a cultural js thing that's developed over time that's not really well communicated

SlayerOrnstein: still missing something because now there's no archwings in wikiadata

build/parser.mjs
```diff
@@ -968,9 +978,8 @@ class Parser {
     if (target.EstimatedVaultedDate) {
       item.estimatedVaultDate = this.normalizeOggDate(target.EstimatedVaultedDate);
     }
-    if (target.Vaulted) {
-      item.vaulted = target.Vaulted;
-    }
+
+    item.vaulted = target.vaulted || wikiaItem.vaulted;
```
AyAyEm, line 982: i've started reviewing the data and it seems like this line should merge objects if both are available, we're missing a bunch of 'wikiaThumbnail' entries and other data, also the parent data is all mangled

[itemsOld.json](https://github.com/WFCD/warframe-items/files/13850184/itemsOld.json)
[itemsNew.json](https://github.com/WFCD/warframe-items/files/13850183/itemsNew.json)
`curl https://termbin.com/7yl0`
`jsondiffpatch /itemsOld.json ./itemsNew.json`
if you guys know of a better solution to diff json let me know

SlayerOrnstein: hmm I'm not entirely sure. Like I said not super versed in js but I don't really see this one would affect wikiaThumbnail (sentence feels a little off but I'm genuinely asking)

AyAyEm: Sorry my mistake i forgot some entries just are kinda random when you build

AyAyEm: and also idk why i thought this line was an object it's a boolean, but still some weird thing is happening and valkyr prime doesn't have the vaulted entry anymore

SlayerOrnstein: That one I will check

AyAyEm:
All the pertinent data for this PR
[itemsOld.json](https://github.com/WFCD/warframe-items/files/13850184/itemsOld.json)
[itemsNew.json](https://github.com/WFCD/warframe-items/files/13850183/itemsNew.json)

[vaultData.json](https://github.com/WFCD/warframe-items/files/13850515/vaultData.json)
[wikia.json](https://github.com/WFCD/warframe-items/files/13850516/wikia.json)

build/parser.mjs
AyAyEm: it's a casing issue i think
```suggestion
    item.vaulted = target.Vaulted || wikiaItem.vaulted;
```

AyAyEm: i rebuild and it didn't fix it uhhh

AyAyEm: oh it actually fixed it my mistaked

SlayerOrnstein: oh well it should be `target.Vaulted` so yeah?

AyAyEm: Yeah all the data from ogg seems to be in pascalcase

build/parser.mjs
```diff
@@ -941,11 +941,21 @@ class Parser {
...
+   * @param {module:warframe-items.Category} category of the data
+   * @param {WikiaData} wikiaData from wikia to apply 
    */
-  addVaultData(item, vaultData) {
+  addVaultData(item, vaultData, category, wikiaData) {
+    let vaultCategory = category;
+    if (item.type === 'Archwing') vaultCategory = 'archwings';
+    if (category === 'Upgrades') vaultCategory = 'mods';
+
+    if (!['weapons', 'warframes', 'mods', 'upgrades', 'archwings'].includes(vaultCategory.toLowerCase())) return;
```
AyAyEm, line 954: Sentinels also have primes
```suggestion
    if (!['weapons', 'warframes', 'mods', 'upgrades', 'archwings', 'sentinels'].includes(vaultCategory.toLowerCase())) return;
```

build/parser.mjs
```diff
@@ -941,11 +941,21 @@ class Parser {
    * Adds releaseDate, vaultDate and estimatedVaultDate to all primes using
    * data from "Ducats or Plat".
    * @param {Item} item data to append vault data to
+
```
AyAyEm, line 944: Unnecessary blank line

AyAyEm:
is it possible to get the vaulted status from relics from wikia? if so would you be willing to do the work in this pr @SlayerOrnstein ? (you've already done alot this is totally optional) if so we could fix #496

SlayerOrnstein:
Hmm it would be but I think that's a feature better suited for warframe-relics then add that into items

build/parser.mjs
SlayerOrnstein: Adding sentinels means adding a scraper for companions (https://warframe.fandom.com/wiki/Module:Companions/data). I can add this but I'm not sure if this would break anything else that would include companions.

SlayerOrnstein: also looks like they also already have it https://api.warframestat.us/items/search/Carrier%20Prime/

AyAyEm: Well this pr is removing the data that sentinels already have, so maybe just use the ogg for it and not the wikia? or also add the  wikia data?

build/parser.mjs
```diff
@@ -968,9 +977,8 @@ class Parser {
     if (target.EstimatedVaultedDate) {
       item.estimatedVaultDate = this.normalizeOggDate(target.EstimatedVaultedDate);
     }
-    if (target.Vaulted) {
-      item.vaulted = target.Vaulted;
-    }
+
+    item.vaulted = target.Vaulted || wikiaItem.vaulted;
```
AyAyEm, line 981: Just realized something, this shouldn't be a boolean or, because it's fine if it's false
```suggestion
    item.vaulted = typeof target.Vaulted === 'boolean' ? target.Vaulted : wikiaItem.vaulted;
```

build/parser.mjs
AyAyEm: Just diff the data that i provided and you'll see the problem
`jsondiffpatch ./itemsOld.json ./itemsNew.json`

build/parser.mjs
```diff
@@ -942,10 +942,19 @@ class Parser {
...
+    if (item.type === 'Archwing') vaultCategory = 'archwings';
+    if (category === 'Upgrades') vaultCategory = 'mods';
+
+    if (!['weapons', 'warframes', 'mods', 'upgrades', 'archwings'].includes(vaultCategory.toLowerCase())) return;
     if (!item.name.endsWith('Prime')) return;
+
     const target = vaultData.find((i) => i.Name.toLowerCase() === item.name.toLowerCase());
+    const wikiaItem = wikiaData[vaultCategory.toLowerCase()].filter((i) => i).find((i) => i.name === item.name);
 
     if (!target) {
```
AyAyEm, line 959: @TobiTenno should wikiaItem be added to this test? So it'll fallback to wikia data in case the ogg data doesn't exist?
```suggestion
    if (!target && !wikiaItem) {
```

SlayerOrnstein: So wikia doesn't have `EstimatedVaultedDate` or `VaultedDate` so if it doesn't exist it would be better to just skip over it since there's no other way to add it from wikia?

build/parser.mjs
SlayerOrnstein: unless the ogg data is wrong or the wikiadata is wrong? that was I figured would happen in either case. if both are false then that's fine

build/parser.mjs
AyAyEm: Well if it's skipped it won't add the vaulted state right?

build/parser.mjs
AyAyEm: it's because in js undefined is cast to false in a boolean comparison, so if we have the target data and vault is false it'll ignore it and look for the wikia which could be false, true or undefined. If ogg is the default data source it doesn't make sense to ignore it when the vaulted state is false.

build/parser.mjs
SlayerOrnstein: Yup but it also won't error out when target doesn't exist but Wikia does

build/parser.mjs
SlayerOrnstein: Ah I understand now

AyAyEm: basically we take ogg data by default unless it's not a boolean (undefined) so if it's not we take whatever the wikia gives us undefined, false, true, in the previous scenario we would take true from ogg and if it's not true we would take undefined, true or false from wikia.

build/parser.mjs
AyAyEm: i see you're right we could just add `target?.ReleaseDate` in that case so it won't error out, because i think being vaulted or not is the priority and not the dates.

SlayerOrnstein: oooo I didn't know I could use that in JS, this is null check right?

AyAyEm: yep it just ignores the next dot if it's undefined or null it's called Optional Chaining
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

build/parser.mjs
AyAyEm: Just remembered there's also nullish coalescing operator
```suggestion
    item.vaulted = target.Vaulted ?? wikiaItem.vaulted;
```

build/parser.mjs
TobiTenno: yeah

build/parser.mjs
TobiTenno: this would be better, since `target.Vaulted` could be false, in which case you'd want to use that.

build/scraper.mjs
TobiTenno: still waiting on this to come back

TobiTenno (changes requested):
if  you're gonna add sentinels & archwings, you should upgrade this to a feat, or do those in a separate PR.

SlayerOrnstein:
> if you're gonna add sentinels & archwings, you should upgrade this to a feat, or do those in a separate PR.

Gonna have to be a feature then because without archwings building breaks and without sentinels data is removed

build/scraper.mjs
SlayerOrnstein: gimme a few lost my gpg keys

TobiTenno:
> > if you're gonna add sentinels & archwings, you should upgrade this to a feat, or do those in a separate PR.
> 
> Gonna have to be a feature then because without archwings building breaks and without sentinels data is removed

aren't both of those only because you added those to being checked in the wiki updates?

those both seem like things that should be in their own features...

SlayerOrnstein:
> > > if you're gonna add sentinels & archwings, you should upgrade this to a feat, or do those in a separate PR.
> > 
> > 
> > Gonna have to be a feature then because without archwings building breaks and without sentinels data is removed
> 
> aren't both of those only because you added those to being checked in the wiki updates?
> 
> those both seem like things that should be in their own features...

Because Warframes and Archwings are under the same category it would try to look for Odonata Prime under warframes. I can if you want have it ignore Archwings in this PR and then add it back in a different PR with sentinels but I guess because I still have to check the category before calling `wikiaItem` I would have to ignore sentinels and possible still end up removing vault fields from sentinels

SlayerOrnstein:
So I'll make this one depend on #506

build/scraper.mjs
TobiTenno: :fearful:

SlayerOrnstein: Line should be back now? Also I thought I did, forgot I put the private and public key in the yubi

build/parser.mjs
```diff
@@ -959,18 +968,18 @@ class Parser {
...
-    if (target.EstimatedVaultedDate) {
+    if (target?.EstimatedVaultedDate) {
       item.estimatedVaultDate = this.normalizeOggDate(target.EstimatedVaultedDate);
     }
-    if (target.Vaulted) {
-      item.vaulted = target.Vaulted;
-    }
+
+    item.vaulted = target?.Vaulted ?? wikiaItem.vaulted;
+    if (item.vaulted === 'N/A') item.vaulted = true;
```
AyAyEm, line 982: Could you explain this line? Like in which case it would be N/A and why?

AyAyEm: i don't like much having this one line of code for one warframe but i guess there's no other way around it

TobiTenno: theoretically would also match other stuff if they ever move Vaulted to a field on item

SlayerOrnstein: I would imagine for primes that come from baro since technically they aren't vaulted but also not unvaulted either

TobiTenno:
i see 3 things that should not be in the PR

TobiTenno (approved):

AyAyEm (approved):
Looks good to me
[vaulted.csv](https://github.com/WFCD/warframe-items/files/14132711/vaulted.csv)

PR #506, opened 2024-01-08
feat: scrape archwings and companions from wikia

SlayerOrnstein:
### What did you fix? 
Doesn't fix anything just lets scarper pull information about archwings and companions. Not sure how to do everything in js so I only used the bare minimum for companions. As Archwings share a lot of information from warframes I used the same transformer.

### Considerations
- Is is a bug fix, feature request, or enhancement? **Feature**


TobiTenno (approved):
much cleaner this way 👍

PR #507, opened 2024-01-08
fix: isPrime property in some items

AyAyEm:
### What did you fix? 
Fix some items not being marked as `isPrime` while other being marked despite not being prime, for example the `Proboscis Cernos`

### Considerations


build/parser.mjs
```diff
@@ -827,7 +827,10 @@ class Parser {
     )
       return;
 
-    item.isPrime = item.uniqueName.includes('Prime');
+    const isPrimeEquipment = item.name.endsWith('Prime') || item.name.startsWith('Prime');
+    const isPrimeMod = item.category === 'Mods' && item.name.startsWith('Primed');
+
+    item.isPrime = isPrimeEquipment || isPrimeMod;
```
AyAyEm, line 833: this also could be like this:
```suggestion
    item.isPrime = item.name.includes('Prime');
```

TobiTenno: nah, i purposely didn't do this, lol. there's items with prime, like `primed chamber` that are definitely _not_ prime.

there's a reason this wasn't how we did this. I promise i think through these things

TobiTenno (changes requested):

build/parser.mjs
TobiTenno: (ironically, i believe i also overrode primed chamber specifically so it would work). i think a category check should probably be used, but i can't remember if i had more context at some point.

AyAyEm: Oh right uhmm, there's a bunch of entries that should be added and some should be excluded

AyAyEm: primed chamber seems to be an odd mod, it can be an edge case but uhmmm, i will give some thought on how we could solve this

AyAyEm: Here it's the data i guatered

[isPrime.csv](https://github.com/WFCD/warframe-items/files/13862905/isPrime.csv)

TobiTenno: ooo, i didn't!

AyAyEm: yeah i just thought going the simple way and not having a bunch of weird conditional, but i think weird is part of warframe-items it makes part of it's beauty

AyAyEm:
Let's summarize the needs and possibilities

- Most primes has a non prime counterpart, and some just modifies slightly the uniqueName
- All primes have prime in their uniqueName although the location varies
- Some items also have prime in their uniqueName

any more items you can think about?

build/parser.mjs
SlayerOrnstein: Could just use `toLowerCase`, would make it so that even if there's a casing change it would still all be lower case. 

Also as for primed mods I never saw them the same as Prime warframes/weapons. I always thought they were just mods that spent too much time in the void lore wise. A way for DE to upgrade needed mods without introducing more mods

AyAyEm:
Right out of the oven full of a bunch of conditionals i even included a switch for the refined tastes

TobiTenno:
I haven't checked the diff on data yet. Will do shortly tonight.

AyAyEm:
no need to be hasty about this as this is just a minor change to the data

TobiTenno (approved):

TobiTenno:
a bunch of these in the diff seem like pointless upgrades, especially the unreleased "prime" mods.

TobiTenno:
@AyAyEm rebase when you can (i don't wanna mess up the signatures)

AyAyEm:
> a bunch of these in the diff seem like pointless upgrades, especially the unreleased "prime" mods.

opsie i included the file where i was trying to see if it was possible to not have the legendary mod check, or you're talking about something else?

TobiTenno:
i'm dumb, i should've stopped this cause you changed the function name and i completely missed it. i'm switching it back. it purposely used the language that was there before cause of how we're operating with it

PR #552, opened 2024-04-14
fix: add regex for melee items that are just skins

SlayerOrnstein:
### What did you fix? 
Makes the melee item type a regex so it doesn't match against melee skins

---

### Reproduction steps

---

### Considerations


TobiTenno:
This has no changes

SlayerOrnstein:
Hold off on this one because I'm still trying to remember where I was with this

PR #561, opened 2024-05-04
fix: implement Items.map so 'length' stays the same

Sainan:
Fixed the same issue as what you apparently already fixed for 'filter'.


Sainan:
Why did you close this PR? The issue is still present on main.

PR #563, opened 2024-05-04
fix: typo causing test failure

Sainan:


Sainan:
Umm, why did you close this?

SlayerOrnstein:
I think gh just auto closes when it does a force push...

SlayerOrnstein:
This should work

Sainan:
Well, all that needs to be done is to write 'Vazarin' instead of 'Varazin'.

SlayerOrnstein:
yeah yeah just push an empty commit on the same branch and it should let you reopen

Issue #566, opened 2024-05-06
'map' does not work as expected on items array


Sainan:
Using 'map' on the items array causes every item to be present in the result, an easy way to see this is by querying 'length':
```JS
const items = new Items({ category: ['Arch-Gun'] });
console.log(items.length);
console.log(items.map((x) => x).length);
```

I've attempted to fix this in #561 by reimplementing the function as was previously done with 'filter', however this fix was seemingly rejected.

Another possible fix would be to not store properties on a class extending Array.

TobiTenno:
closed by #567

Issue #562, opened 2024-05-04
Can't checkout latest commit on Windows


Sainan:
```
error: invalid path 'data/img/gauss-prime-theme"redline".png'
```

TobiTenno:
a better way to phrase this would be "illegal filename character `"` in image names & paths". this is being handled.

PR #567, opened 2024-05-06
fix: override `Array.prototype.map`

SlayerOrnstein:
### What did you fix? 
Same fix as #566 but @Sainan is unable to rebase due to #562 

---

### Reproduction steps

---

### Considerations


Sainan:
Just FYI, you can merge a PR even if it's not rebased.

TobiTenno:
we're aware, but you were missing half the implementation

Sainan:
I'm not sure what you mean, I tested locally with `npm install ../warframe-items` and the map function was working as expected.

Sainan:
Nevermind, I see it was also added to the .mjs file here...

So much overengineering here yet still having code duplication like that? Wow.

Sainan:
I think I was being quite respectful and professional.

Now I am just mocking you because you clearly had a very big ego about how overengineered this package is to the point were you actually banned when I suggested it should be simpler so it would work on Windows without problems, for example. 😀

SlayerOrnstein:
So completely unaware, got it. So I'm just gonna say this, if you're going to contribute follow the rules set by the repo not "you can merge without rebasing" if you don't want to that's fine you also don't have to contribute 

That is the end of this whole conversion, you were told what needed to be told and you refused, anything after this as harsh as it seems falls under harassment

Sainan:
Harass deez

Sainan:
Like, how the fuck can you equate me reporting issues to "harassment"? You people should not be doing software development if your ego and emotions are this unstable.

build/parser.mjs
```diff
@@ -532,6 +532,7 @@ class Parser {
         .replace('/', '')
         .replace(/[ /*]/g, '-')
         .replace(/[:<>[\]?!]/g, '')
+        .replaceAll('"','')
```
TobiTenno, line 535:
```suggestion
        .replace(/[:<>[\]?!"]/g, '')
```

TobiTenno (approved):

PR #594, opened 2024-07-30
fix: use last part of resource name to avoid image name collision

SlayerOrnstein:
### What did you fix? 
Fixed a collision that can happen when two items share the same name by using the last part of item's unique name instead of it's in game name.

It shouldn't cause any issues with people who get the image name from the `warframe-item` data but if you're pulling the image another way then maybe

image

---

### Reproduction steps

---

image
image

### Considerations


TobiTenno:
could you preserve the dash behavior and separate on the caps within the last part of the unique name?

SlayerOrnstein:
> could you preserve the dash behavior and separate on the caps within the last part of the unique name?

That should do it

TobiTenno (approved):

PR #635, opened 2024-10-15
fix: introduce a partial hash to identify images

SlayerOrnstein:
### What did you fix? 
Main idea is to try and fix #594 and #599. Turns out that even unique names aren't completely unique as the last part of it's resource name can be the same so long as the resource path is different i.e.

```
/Lotus/Upgrades/Mods/Rifle/Beginner/WeaponAmmoMaxModBeginner
/Lotus/Upgrades/Mods/Shotgun/Beginner/WeaponAmmoMaxModBeginner
/Lotus/Upgrades/Mods/Pistol/Beginner/WeaponAmmoMaxModBeginner
```

So this implements the same idea as #632 by generating a SHA-256 on the unique name and appending a partial part of it as an identifier to avoid naming collisions (does not apply to shared images)

> [!WARNING]
> During my testing I found that Jade was not the only image being overwritten by items that shared the same name. For example Archon Flow has two entries with different images
> ```
> /Lotus/Upgrades/Mods/Warframe/Kahl/KahlAvatarPowerMaxMod
> /Lotus/Upgrades/Mods/Warframe/Kahl/KahlAvatarDropEnergySubMod
> ```
> So in this case `KahlAvatarPowerMaxMod` would overwrite `KahlAvatarDropEnergySubMod` but the real image for `KahlAvatarDropEnergySubMod` is the same for [Flow](https://warframe.fandom.com/wiki/Flow)
>
> So if you find an image that doesn't make sense check to see if there's a duplicate item

---

### Reproduction steps

---

### Considerations


SlayerOrnstein:
[`previous`](https://github.com/WFCD/warframe-items/pull/635/commits/95f44e08aa021a38bdfa30ec3faf93469c97170d) appended the `uniqueName` when the previous item and current item shared the same name, since the SHA is acting as an identifier even if the previous item shares the same name the `imageName` wouldn't be the same because of the appended the SHA

That is how I read the code at least

TobiTenno (approved):

PR #681, opened 2025-01-04
feat: add `uniqueName` to component relic drops

SlayerOrnstein:
### What did you fix? 
Add the relic's `uniqueName` as a pseudo link in order to make it easier to pull or link the full relics's information

---

### Reproduction steps

---

### Considerations
- Is is a bug fix, feature request, or enhancement? **Feature**

## Summary by CodeRabbit

- **New Features**
	- Enhanced relic drop detection and processing logic
	- Added optional unique name support for drops

- **Refactor**
	- Improved relic data handling method for better clarity and maintainability

### DIFF
--- a/build/parser.mjs
+++ b/build/parser.mjs
@@ -1048,8 +1048,9 @@ class Parser {
       // Relic names are /(Lith|Meso|Neo|Axi|Requiem) (\w+)/
       const related = relics.filter((relic) => relic.name.toLowerCase() === relicName);
 
-      if (link) [relicItem.uniqueName] = Array.from(new Set(related.map((relic) => relic.uniqueName).flat()));
-      else {
+      if (link) {
+        [relicItem.uniqueName] = Array.from(new Set(related.map((relic) => relic.uniqueName).flat()));
+      } else {
         relicItem.locations = Array.from(new Set(related.map((relic) => relic.locations).flat()));
         relicItem.rewards = Array.from(new Set(related.map((relic) => relic.rewards).flat()));
         [relicItem.marketInfo] = Array.from(new Set(related.map((relic) => relic.warframeMarket)));
--- a/build/parser.mjs
+++ b/build/parser.mjs
@@ -1038,20 +1038,33 @@ class Parser {
    * @param {Array<module:@wfcd/relics.TitaniaRelic>} relics relic array to search
    */
   addRelics(item, relics) {
-    if (item.type !== 'Relic') return;
+    const hasRelicDrop = item.components?.some((c) => c.drops?.some((d) => d.location?.includes('Relic')));
+    if (item.type !== 'Relic' && !hasRelicDrop) return;
 
-    // Item names are /(Lith|Meso|Neo|Axi|Requiem) (Relic|\w+ \w+)/
-    const relicName = item.name.toLowerCase().split(' ').slice(0, 2).join(' ');
+    const addRelic = (relicItem, name, link = false) => {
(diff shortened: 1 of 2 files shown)


TobiTenno (changes requested):
at present, this inflates the size of relic drops too much, too large of a data file increase :-1: 

if we want to write something that "merges" it in at instantiation time and people are willing to take the overhead, that's fine, but i don't want to load this much into memory when people load items

SlayerOrnstein:
I could switch it to just use the unique name instead. This way it'll be easier to pull the relic directly

TobiTenno:
> 47M to 48M if `type` is changed from the component drop name to `Relic`

I think that would break the usability of the type field, as it's currently parsed out as the component drop name...

SlayerOrnstein:
@TobiTenno yup, I considered it but I also had the same argument against it, I only mentioned it because I wanted to be transparent with different options to try and maximize the information for component drops

SlayerOrnstein:
@TobiTenno So realized a few things

1. The original size was not 47M it was 46M
2. `mastrable` was being added to every item (fixed now and it reduced the size)
3. With just the `uniqueName` and `vaulted` the size final size is at 47M

build/parser.mjs
```diff
@@ -1070,12 +1095,16 @@ class Parser {
...
+    if (item.parent) return;
     item.masterable = masterableCategories.categories.includes(item.category);
 
     if (item.type?.includes('Component') || item.category === 'Pets') {
       const regex = new RegExp(masterableCategories.regex);
       item.masterable = regex.test(item.uniqueName);
     }
+
+    // Assume the item is outside the masterable category and remove it completely
+    if (!item.masterable) delete item.masterable;
```
TobiTenno, line 1107: i thought we wanted to keep these booleans, not just have them not exist

SlayerOrnstein: Removing it when it's false would reduce data bloat as outside of things that won't give XP it seems a bit excessive.

For example if I were to pull a resource item I wouldn't need to check if it's masterable because resources don't give mastery points

having said that I can see why it would be something to have for maybe Zaws/K-Guns/K-drive and nodes?

TobiTenno: yeah, that's why i'd previously stopped deleting fields, at the request of others to make sure if it's possible on the type it stays on the type, since some languages don't consume fields missing nicely

SlayerOrnstein: fair enough

build/parser.mjs
```diff
@@ -1038,20 +1038,33 @@ class Parser {
...
-    item.locations = Array.from(new Set(related.map((relic) => relic.locations).flat()));
-    item.rewards = Array.from(new Set(related.map((relic) => relic.rewards).flat()));
-    [item.marketInfo] = Array.from(new Set(related.map((relic) => relic.warframeMarket)));
+      // Relic names are /(Lith|Meso|Neo|Axi|Requiem) (\w+)/
+      const related = relics.filter((relic) => relic.name.toLowerCase() === relicName);
 
-    const [vaultInfo] = Array.from(new Set(related.map((relic) => relic.vaultInfo)));
-    item.vaulted = vaultInfo?.vaulted;
-    item.vaultDate = vaultInfo?.vaultDate;
+      if (link) [relicItem.uniqueName] = Array.from(new Set(related.map((relic) => relic.uniqueName).flat()));
```
TobiTenno, line 1051: i don't really like this format, just cause it's a little... weird of a format and easily prone to misunderstanding. let's make it more verbose so we don't have as weird of a format.

SlayerOrnstein: Fair point 😅

PR #718, opened 2025-03-30
feat: expose 'wikiAvailable', match on uniqueName

### LINKED ISSUE #600: Feature Request: `obtainable` flag for items
**Is your feature request related to a problem? Please describe.**
Currently, it does not seem possible to tell which items are actually obtainable. An example of this being "Primed" Ammo Drum — it may exist internally, but it is completely unobtainable currently and displaying it for my own purposes would not be wise.

**Describe the solution you'd like**
I would like to see an `obtainable` flag on items, in order to distinguish which items can actually be obtained.

Currently, all I can think of is lazily parsing the "uniqueName" of items and trying to figure out which items don't exist based on that. However, this does not feel like an ideal solution in the slightest.

**Additional context**
<https://api.warframestat.us/items/search/Ammo%20Drum/>
image

TobiTenno:
### What did you fix? 
closes #600

---

### Reproduction steps

---

### Considerations
- Does this contain a new dependency? **[Yes]** but for dev only

## Summary by CodeRabbit

- **New Features**
  - Enhanced item metadata processing now uses unique identifiers for improved matching and clearly flags when wiki data is available.
  
- **Chores**
  - Introduced a JSON diffing library to support internal data comparisons.
  
- **Tests**
  - Updated test logging now provides clearer comparisons for verifying item data consistency.
  
- **Documentation**
  - Updated interfaces to include a new property indicating wiki availability and adjusted related properties accordingly.

### DIFF
--- a/build/parser.mjs
+++ b/build/parser.mjs
@@ -912,8 +912,11 @@ class Parser {
     if (item.category === 'Archwing') wikiCategory = 'archwings';
     if (category === 'Sentinels') wikiCategory = 'companions';
 
-    const wikiaItem = wikiaData[wikiCategory].filter((i) => i).find((i) => i.name === item.name);
+    const wikiaItem = wikiaData[wikiCategory]
+      .filter((i) => i)
+      .find((i) => i.uniqueName === item.uniqueName);
     if (!wikiaItem) return;
+    item.wikiAvailable = true;
 
     switch (category.toLowerCase()) {
       case 'sentinels':
--- a/build/parser.mjs
+++ b/build/parser.mjs
@@ -912,9 +912,7 @@ class Parser {
     if (item.category === 'Archwing') wikiCategory = 'archwings';
     if (category === 'Sentinels') wikiCategory = 'companions';
 
-    const wikiaItem = wikiaData[wikiCategory]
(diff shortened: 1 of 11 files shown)


TobiTenno:
@coderabbitai see https://github.com/WFCD/warframe-items/issues/600#issuecomment-2764758648 for how this applies to #600

SlayerOrnstein (approved):
LGTM

Cypheriel:
Perhaps this goes without saying, but I'm not so sure that this *really* solves the original issue. As with anything Wiki-related, there's going to be a concern of reliability with the data we're working with. Depending on the use case, this could hamper external services relying on this API. However, in addition, there may be a case where a page exists for an item, an item technically exists, and is not obtainable nor tradable. Also, I worry some items may not directly translate well to a Wiki URL.
 I would say there should be an analysis of the data before marking the original issue as completed.
 
TL; DR: The original issue and this solution may provide the same information at the moment or any given time, but technically represent two different ideas and uses.

TobiTenno:
I realize that, but there's literally no non-manual way of maintaining "is this obtainable" that isn't an inference based on data outside DE's manifest

PR #753, opened 2025-06-25
fix: check `wikia.weapons` type before falling back to type.name

### LINKED ISSUE #752: :rocket: fall back to wikiData.Class instead of Rifle
Default to `wikiData.Class` instead of the current `Weapon` -> `Rifle` matching that defaults for a lot of weapon types

please make the item type matcher be capable of pulling data from the wiki Class, as that seems to be more expansive and verbose compared to our `type` matcher for some weapons that don't have clear ways of knowing what `type` various weapons should be

Looked at adding a uniqueName matcher for the affected weapon (specifically Lanka being marked as "Rifle" and not "Sniper" or "Sniper Rifle"), but there's insufficient information to do so.
Additionally looked at other metadata, nothing was sufficient for our needs.

**Additional context**

justsomelegs:
### What did you fix? 
check `wikia.weapons` type before falling back to `type.name`
closes #752 

---

 You can see my fix [on this line](#diff-4288e98278b6f4723cd3f212f07c2c17197f089125f10a50eae56c2ac7a3a7c1R45)

### Considerations


build/parser.mjs
```diff
@@ -445,12 +445,13 @@ class Parser {
...
       // eslint-disable-next-line no-restricted-syntax
       for (const type of types) {
+        const weapon = data.wikia.weapons.find((entry) => entry.name === item.name);
         const contains = type.regex
           ? new RegExp(type.id, 'ig').test(item.uniqueName)
           : item.uniqueName.includes(type.id);
         if (contains) {
           if (type.append) item.type = `${item.type}${type.name}`;
-          else item.type = type.name;
+          else item.type = weapon?.type ?? type.name;
```
TobiTenno: while this does override, this doesn't fall back, which is what i mentioned on the issue.

this needs to have a new type resolution entirely in the json of "wiki" instead of just using the wiki one. so `type` needs to contain enough info that you can flag off of in this so that instead of directly going to `Rifle` ever, it falls back to `weapon?.type`

justsomelegs: So something like this?:

``` js

if (type.wiki && weapon?.type) {
	const weapon = data.wikia.weapons.find((entry) => entry.name === item.name);
	item.type = weapon.type;
} else {
	item.type = type.name;
}
```

with the JSON looking like this?:

```js
{
    "id": "Primary",
    "name": "Rifle",
    "wiki": true
  },
```

justsomelegs: or even?:

```js
if (type.append) item.type = `${item.type}${type.name}`;
if (item.wiki) item.type = data.wikia.weapons.find((entry) => entry.name === item.name)?.type ?? type.name;
else {
	item.type = type.name;
}
```

TobiTenno:
```ts
else if (type.wiki)
else item.type = type.name;
```
otherwise the third one would probably be the preference

config/itemTypes.json
```diff
@@ -937,28 +937,32 @@
   },
   {
     "id": "/LongGuns",
-    "name": "Rifle"
+    "name": "Rifle",
+    "wiki": true
```
TobiTenno: nonononononono

TobiTenno (changes requested):
every generic type should **not** default to the wiki, just the one instance we're seeing. targeted fixes, not wide-spread blanket stuff.

justsomelegs:
Is this still too wide of a fix?

build/parser.mjs
```diff
@@ -450,7 +450,11 @@ class Parser {
           : item.uniqueName.includes(type.id);
         if (contains) {
           if (type.append) item.type = `${item.type}${type.name}`;
-          else item.type = type.name;
+          else if (item.wiki)
+            item.type = data.wikia.weapons.find((entry) => entry.name === item.name)?.type ?? type.name;
+          else {
+            item.type = type.name;
+          }
```
TobiTenno, line 457: could you add some tests around known examples of this so that we can verify the scope of the change?

PR #830, opened 2025-12-30
fix: add missing images

AyAyEm:
### What did you fix? 
Add missing images

### Considerations
- Is is a bug fix, feature request, or enhancement? **[Maintenance]**

For some reason when building it also update many entries of itemCount and parents in the All.json file. We should probably change the code to check if all the entries in the cache are in fact stored in disk.

## Summary by CodeRabbit

* **Chores**
  * Refreshed cache handling for a specific item to ensure consistency
  * Improved build process reliability by ensuring cache updates occur exclusively after successful image processing and optimization, preventing potential data inconsistencies from interrupted or failed operations

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/build/build.mjs
+++ b/build/build.mjs
@@ -261,8 +261,6 @@ class Build {
         };
         const image = await get(imageUrl).catch(retry).catch(retry);
 
-        this.updateCache(item, cached, hash, isComponent);
-
         await sharp(image).toFile(filePath);
         await minify([filePath], {
           destination: basePath,
@@ -273,6 +271,8 @@ class Build {
             }),
           ],
         });
+
+        this.updateCache(item, cached, hash, isComponent);
       } catch (e) {
         // swallow error
         console.error(e);


AyAyEm:
It seems there's a bunch of images missing i'm building it now, up to 168

AyAyEm:
There's more problems with the images than i thought, there might need one or two prs to fix related issues.

test/index.spec.mjs
```diff
@@ -1,8 +1,11 @@
 import assert from 'node:assert';
+import fs from 'node:fs/promises';
+import nodePath from 'path';
 import { resolve } from 'node:path';
```
TobiTenno: yeah, you should import `join`, not use a differently scoped path, thanks

build/build.mjs
```diff
@@ -176,7 +176,7 @@ class Build {
     // No need to go through every item if the manifest didn't change. I'm
     // guessing the `fileTime` key in each element works more or less like a
     // hash, so any change to that changes the hash of the full thing.
-    if (!hashManager.hasChanged('Manifest')) return;
+    // if (!hashManager.hasChanged('Manifest')) return;
```
TobiTenno: this should definitely not be removed

build/build.mjs
```diff
@@ -273,6 +271,8 @@ class Build {
             }),
           ],
         });
+
+        this.updateCache(item, cached, hash, isComponent);
```
TobiTenno, line 275: doesn't seem like this actually does anything. if you'd like to explain your intention, feel free, but none of the params are modified by moving this

AyAyEm: sometimes the cache is updated before saving the file, then image is missing in the next runs. for some reason sharp generates an error on some images some of the time, so only updating the cache when it's saved to disk is better

build/build.mjs
AyAyEm: i agree but for testing purposes it's better this way, im using the draft as a way to fast commit the changes im making even if theyre not going to be there at the end

test/index.spec.mjs
AyAyEm: not sure it'll stay like that

AyAyEm:
This draft is still in design phase, so no need for the code be compliant with lint or to be well designed. Remember this is basically me testing around on what we can do about the problem. I could very much discard all changes.

test/index.spec.mjs
AyAyEm: also importing join is not that good as join by itself doesnt have much meaning. Join arrays, strings, objects, everything? path.join is better because you know what it's joining.

AyAyEm:
I'll just add the images that are missing first and then we'll think about what to do with the other issues.

test/index.spec.mjs
TobiTenno: that's your opinion, and given the pattern that is already in place (see the immediate following line) you should continue the pattern in place.

PR #858, opened 2026-02-12
docs(tradable): explain component check

Clarity-89:
And a comment to clarify how the tradable status is computed.

### Considerations

## Summary by CodeRabbit

* **Documentation**
  * Added clarification on tradability restrictions for Prime items and trading behavior for components.

### DIFF
--- a/build/tradable.mjs
+++ b/build/tradable.mjs
@@ -10,7 +10,6 @@ const builtUntradable = [
   'Launcher',
   'Sniper',
 ];
-/* eslint-disable-next-line valid-jsdoc */
 /**
  * Gate built Prime items (Warframes, weapons) from being marked tradable.
  * Built Prime items are not directly tradable -- only their individual
@@ -23,6 +22,9 @@ const builtUntradable = [
  * Item-specific tradability overrides live in config/overrides.json
  * (keyed by uniqueName) and are applied by parser.mjs applyOverrides()
  * after this check runs.
+ *
+ * @param {module:warframe-items.Item} item Item to check
+ * @returns {boolean}
  */
 const tradableConditions = (item) => !(builtUntradable.includes(item.type) && item.name.match(/Prime/gi));
 


build/tradable.mjs
```diff
@@ -10,7 +10,32 @@ const builtUntradable = [
...
 ];
-const tradableConditions = (item) => !(builtUntradable.includes(item.type) && item.name.match(/Prime/gi));
+// Explicit tradability overrides for built Prime items that the component check gets wrong.
+// true = tradable despite no components (e.g. Baro Ki'teer weapons)
+// false = untradable despite having components (e.g. quest-locked weapons)
+const builtPrimeOverrides = {
+  'Gotva Prime': true,
+  'Galariak Prime': false,
+  'Sagek Prime': false,
+};
```
TobiTenno: If you're making an override, move it out of code and into a config file, and use the unique names, not the localized name.

test/tradable.spec.mjs
```diff
@@ -0,0 +1,199 @@
...
+      { name: 'Barrel', tradable: true },
+      { name: 'Receiver', tradable: true },
+      { name: 'Orokin Cell', tradable: false },
+    ],
+  });
+
+describe('tradable', () => {
+  describe('standard relic-farmed Prime items', () => {
+    it('should mark Prime warframes with tradable components as tradable', () => {
+      const item = mockPrimeWithComponents('Ash Prime', 'Warframe');
```
TobiTenno: this should not be tradable. it's parts are, not the warframe itself, much like the rest in this first list. no built prime is tradable.

Clarity-89: But what about sets, how do we mark those as tradable?

TobiTenno: "Sets" aren't an item, they're a collection of items, which are individually tradable.

build/tradable.mjs
TobiTenno: this is also not exhaustive of all baro prime weapons, and should probably use something on the item, like either a drop source or add that it's a baro weapon in the weapon overrides, and key off of that, instead of making an explicit list of tradable weapons, because that requires an extra PR to add to this specific list, and adding to configs is a bit more straightforward

test/tradable.spec.mjs
Clarity-89: Hmm, maybe I misunderstood how this is supposed to work then 🤔

TobiTenno: so each tradable boolean is whether that item, built, bought, whatever, is tradable (obviously excluding used items, as those are never tradable in general, like any partially ranked weapons). each prime item generally has a blueprint (like the loki prime blueprint), which would indicate if the blueprint itself is tradable, etc.

Clarity-89: Yeah, it makes sense. Instead of changing the logic for how the tradable status is computed, I added a comment explaining how it works.

TobiTenno: could you change the pr title from fix to docs?

TobiTenno: nvm, i can do that, ty for the documentation.

PR #901, opened 2026-04-10
fix: ensure i18n objects are shallow copied

mateo-leal:
### What did you fix? 
When changing locales within the same runtime environment, the returned array did not include the i18n entries for the new locale.
This pull request replaces the reference to the global i18n object with a shallow copy to resolve this issue.

---

### Reproduction steps

---

### Considerations
- Is a bug fix, feature request, or enhancement? **Bug Fix**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved internationalization data handling to properly isolate locale entries and prevent shared reference mutations.

* **Tests**
  * Added test coverage for multi-locale internationalization scenarios, validating correct behavior across sequential locale operations.

### DIFF
--- a/index.js
+++ b/index.js
@@ -105,7 +105,8 @@ module.exports = class Items extends Array {
         if (this.options.i18n) {
           // only insert i18n for the objects we're inserting, so we don't bloat memory
           if (Array.isArray(this.options.i18n)) {
-            const raw = i18n[item.uniqueName];
+            const itemI18n = i18n[item.uniqueName];
+            const raw = itemI18n ? { ...itemI18n } : undefined;
             // only process if passed language is a supported i18n value
             if (raw) {
               Object.keys(raw).forEach((locale) => {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -105,7 +105,8 @@ module.exports = class Items extends Array {
         if (this.options.i18n) {
           // only insert i18n for the objects we're inserting, so we don't bloat memory
           if (Array.isArray(this.options.i18n)) {
-            const raw = i18n[item.uniqueName];
+            const itemI18n = i18n[item.uniqueName];
+            const raw = itemI18n ? { ...itemI18n } : undefined;
```
AyAyEm, line 109: What other value besides an object or undefined itemI18n might be so it needs this check?

mateo-leal: None, but without that verification the branch coverage percentage requires a test that cannot be done on the if statement of line 111

TobiTenno (approved):

PR #993, opened 2026-09-05
fix: scraper for necramech, wiki exception Dark Split-Sword

### LINKED ISSUE #968: [Bug] Missing "wikiaUrl" data for 3 items
### Describe the bug

In the provided /data/jsons/ on the github a few items are missing thier wikiaUrls

### NodeJS Version

NA

### npm Version

NA

### @wfcd/items Version

v1.1275.35

### Access Method

github download

### What happened?

Three items from the data/json do not have the "wikiaUrl" field but they do have entries on the wiki:

**- Warframes.json**
- Bonewidow
- Voidrig

**- Melee.json**
- Dark Split-Sword

not sure whats causing it, but I found this because I was trying to get the wiki url for all the masterable items and my thing was breaking because these ones were missing. not really a big deal as i can just add in exceptions for these but still an issue nonetheless.

Lugia00-boop:
### What did you fix?
closes #968 

---

### Reproduction steps

---

See new necramech scraper and changes made in 
transformWeapon.ts [204-210 and [230]

### Considerations

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Necramech collection with metadata, images, wiki links, configuration details, and market values.
  * Added support for Voidrig and Bonewidow entries.

* **Bug Fixes**
  * Corrected the displayed name for Dark Split-Sword (Heavy Blade) to “Dark Split-Sword.”
  * Improved data validation so Dark Split-Sword, Voidrig, and Bonewidow are correctly recognized as available on the wiki.

### DIFF
--- a/build/scraper.ts
+++ b/build/scraper.ts
@@ -16,7 +16,7 @@ import VersionScraper from './wikia/scrapers/VersionScraper';
 import readJson from './readJson';
 import sleep from './sleep';
 import { get, getJSON, retryAttempts } from './network';
-import type { WikiaData, WikiaWarframe } from './types/shared';
+import type { WikiaData } from './types/shared';
 import type { TitaniaRelic } from '@wfcd/relics';
 import type { Patchlogs } from '@wfcd/patchlogs';
 
@@ -291,6 +291,7 @@ class Scraper {
     await sleep(100);
     const warframes = await new WarframeScraper().scrape();
     bar.tick();
+    await sleep(100);
     const necramechs = await new NecramechScraper().scrape();
     bar.tick();
     await sleep(100);
@@ -312,29 +313,10 @@ class Scraper {
     const vaultData = await new VaultScraper().scrape();
     bar.tick();
 
-    // Combine scraped warframe and necramech data to keep
-    // current data format (mechs and frames are 1 thing)
-    // for compatibility reasons
-    necramechs.forEach((mech) => {
-      const mechToFrame: WikiaWarframe = {
-        name: mech.name,
-        uniqueName: mech.uniqueName,
-        url: mech.url,
-        conclave: mech.conclave,
-        mr: mech.mr,
-        polarities: mech.polarities,
-        sprint: mech.sprint,
-        introduced: mech.introduced,
-        vaulted: mech.vaulted,
-        thumbnail: mech.thumbnail,
-        marketCost: mech.marketCost,
-      };
-      warframes.push(mechToFrame);
-    });
-
     return {
       weapons,
       warframes,
+      necramechs,
       mods,
       versions,
       ducats,
--- a/build/scraper.ts
+++ b/build/scraper.ts
@@ -10,12 +10,13 @@ import CompanionScraper from './wikia/scrapers/CompanionScraper';
 import ModScraper from './wikia/scrapers/ModScraper';
 import WeaponScraper from './wikia/scrapers/WeaponScraper';
 import WarframeScraper from './wikia/scrapers/WarframeScraper';
+import NecramechScraper from './wikia/scrapers/NecramechScraper';
 import VaultScraper from './wikia/scrapers/VaultScraper';
 import VersionScraper from './wikia/scrapers/VersionScraper';
 import readJson from './readJson';
 import sleep from './sleep';
 import { get, getJSON, retryAttempts } from './network';
-import type { WikiaData } from './types/shared';
+import type { WikiaData, WikiaWarframe } from './types/shared';
 import type { TitaniaRelic } from '@wfcd/relics';
 import type { Patchlogs } from '@wfcd/patchlogs';
 
@@ -290,6 +291,8 @@ class Scraper {
     await sleep(100);
     const warframes = await new WarframeScraper().scrape();
     bar.tick();
+    const necramechs = await new NecramechScraper().scrape();
+    bar.tick();
     await sleep(100);
     const mods = await new ModScraper().scrape();
     bar.tick();
@@ -308,6 +311,14 @@ class Scraper {
     await sleep(100);
     const vaultData = await new VaultScraper().scrape();
     bar.tick();
+
+    // Combine scraped warframe and necramech data to keep 
+    // current data format (mechs and frames are 1 thing)
+    // for compatibility reasons
+    necramechs.forEach(mech => {
+      const mechToFrame = mech as WikiaWarframe
+      warframes.push(mechToFrame)
+    });
 
     return {
       weapons,
--- a/build/scraper.ts
+++ b/build/scraper.ts
@@ -272,7 +272,7 @@ class Scraper {
    * @returns wikia data
    */
   async fetchWikiaData(): Promise<WikiaData> {
-    const bar = new Progress('Fetching Wikia Data', 9);
+    const bar = new Progress('Fetching Wikia Data', 10);
     const ducats: WikiaDucat[] = [];
     const ducatsWikia = await get('https://wiki.warframe.com/w/Ducats/Prices/All', true);
     const $ = load(ducatsWikia as string);
--- a/build/scraper.ts
+++ b/build/scraper.ts
@@ -312,12 +312,24 @@ class Scraper {
     const vaultData = await new VaultScraper().scrape();
     bar.tick();
 
-    // Combine scraped warframe and necramech data to keep 
+    // Combine scraped warframe and necramech data to keep
     // current data format (mechs and frames are 1 thing)
     // for compatibility reasons
-    necramechs.forEach(mech => {
-      const mechToFrame = mech as WikiaWarframe
-      warframes.push(mechToFrame)
+    necramechs.forEach((mech) => {
+      const mechToFrame: WikiaWarframe = {
+        name: mech.name,
+        uniqueName: mech.uniqueName,
+        url: mech.url,
+        conclave: mech.conclave,
+        mr: mech.mr,
+        polarities: mech.polarities,
+        sprint: mech.sprint,
+        introduced: mech.introduced,
+        vaulted: mech.vaulted,
+        thumbnail: mech.thumbnail,
+        marketCost: mech.marketCost,
+      };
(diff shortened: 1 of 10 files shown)


build/scraper.ts
```diff
@@ -291,6 +292,8 @@ class Scraper {
     await sleep(100);
     const warframes = await new WarframeScraper().scrape();
     bar.tick();
+    const necramechs = await new NecramechScraper().scrape();
+    bar.tick();
```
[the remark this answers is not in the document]
Lugia00-boop: missed this before, updated in latest push

test/index.spec.mjs
```diff
@@ -192,6 +192,42 @@ const test = (base) => {
...
+        const matches = data.weapons
+          .filter((i) => i.name === 'Dark Split-Sword')
+          .map((i) => {
+            delete i.patchlogs;
+            return i;
+          });
+        const dd = dedupe(matches);
+        assert.strictEqual(dd.length, matches.length, 'Before and after dedupe should match');
+
+        assert.strictEqual(matches[0].wikiAvailable, true)
```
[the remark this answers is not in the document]
Lugia00-boop: other tests in the integrity are not consistent on this, but it wouldn't hurt, so added.

Lugia00-boop:
Just a note, but I think might  have the same problem as the other #976 where the build tests might fail because the tests will be using json from before changes were made, they did pass when husky ran the commit checks locally for me.

TobiTenno:
the tests use json from the build, not pre-change.... take a look at the CI, that's why the build step has to run first.

test/index.spec.mjs
```diff
@@ -192,6 +192,45 @@ const test = (base) => {
...
     describe('integrity', async () => {
       beforeEach(gc);
+      it('should have wikia available for Dark Split-Sword', async () => {
+        const matches = data.weapons
+          .filter((i) => i.name === 'Dark Split-Sword')
+          .map((i) => {
+            delete i.patchlogs;
+            return i;
+          });
+        const dd = dedupe(matches);
```
TobiTenno, line 202: you shouldn't be deduping if you're enforcing that there should only be 1 match...

test/index.spec.mjs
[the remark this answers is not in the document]
TobiTenno: they don't, though, you deduped the array first...

test/index.spec.mjs
```diff
@@ -192,6 +192,45 @@ const test = (base) => {
...
+        assert.strictEqual(matches[0].wikiAvailable, true)        
+      });
+      it('should only have wikia available for Voidrig', async () => {
+        const matches = data.warframes
+          .filter((i) => i.name === 'Voidrig')
+          .map((i) => {
+            delete i.patchlogs;
+            return i;
+          });
+        const dd = dedupe(matches);
```
TobiTenno, line 215: deduping after filtering seems like the opposite of enforcing it properly.

test/index.spec.mjs
```diff
@@ -192,6 +192,45 @@ const test = (base) => {
...
+        assert.strictEqual(dd.length, matches.length, 'Before and after dedupe should match');
+        assert.strictEqual(matches.length, 1);
+
+        assert.strictEqual(matches[0].wikiAvailable, true)
```
TobiTenno, line 219: not sure how it let your code through like this, but you need semicolons and no trailing spaces

test/index.spec.mjs
```diff
@@ -192,6 +192,45 @@ const test = (base) => {
...
+          .filter((i) => i.name === 'Bonewidow')
+          .map((i) => {
+            delete i.patchlogs;
+            return i;
+          });
+        const dd = dedupe(matches);
+        assert.strictEqual(dd.length, matches.length, 'Before and after dedupe should match');
+        assert.strictEqual(matches.length, 1);
+
+        assert.strictEqual(matches[0].wikiAvailable, true)
```
TobiTenno, line 232: semicolons and trailing spaces

build/scraper.ts
```diff
@@ -310,6 +313,14 @@ class Scraper {
     const vaultData = await new VaultScraper().scrape();
     bar.tick();
 
+    // Combine scraped warframe and necramech data to keep 
+    // current data format (mechs and frames are 1 thing)
+    // for compatibility reasons
+    necramechs.forEach(mech => {
+      const mechToFrame = mech as WikiaWarframe
+      warframes.push(mechToFrame)
+    });
```
TobiTenno, line 322: necramechs should produce necramechs. if it's actually a warframe, produce a warframe type from the necramech scraper, otherwise make necramech extend warframe in the types and add your additional fields. don't blind cast it like this; and please follow the styles of the rest of the project. i'm still not sure why the linting isn't failing on these, but it should be.

TobiTenno (changes requested):

Lugia00-boop:
That makes sense, I may of been misunderstanding the CI process when I looked at it before. But if that's the case it should be fine then for the tests? not sure why the build was failing with my last PR then. 
No idea why my linting isnt flagging on this, but I'll work on the fixes.

TobiTenno:
Looks like style rules went away, def my fault. I'll get it fixed upstream.

build/scraper.ts
Lugia00-boop: Yeah I was going to ask about this, I didn't feel like this was a good way to go about this. My initial thought was to just use wikiaWarframe for the necramech scraper, but i was overthinking a bit and was worried that, if (for whatever reason) necramechs got a new field that warframes didn't have technically that could be an issue as they wouldn't be directly convertable, but if they just stay grouped in with the warframes that field would get dropped anyway, and the solution I did here just effectively does that anyway.

test/index.spec.mjs
Lugia00-boop: If I could get some clarification, from what I was understanding the dedupe first makes sure there are no duplicate entry's that are identical, then the exactly one makes sure that if it happens to somehow match something that has the same name but isn't an exact duplicate (ie won't get deduped). 
it may be a bit heavy on the checks since this test is only checking wikiaUrl existence while the other tests (like the maulson uniqueness check) are explicitly checking for the uniqueness, but i'm not sure how its logically wrong.

build/scraper.ts
TobiTenno: yup, that's exactly why  it should produce necramechs and not WikiaWarframes and let them get picked up to match the right type and get new fields

TobiTenno:
hey, give it a rebase when you get a chance, we got the style enforcement in a bit better. sorry for my rants about that.

Lugia00-boop:
linting is flagging now and I've pushed an update with all the linting fixed and some edits to the necramech parsing.

I updated the necramech to warframe conversion to manually move all the fields from necramechs into a warframe, but I debated for quite a while but couldn't come up with a much better way to parse it that made me happy.
I thought about having necramechs interface being an extension of the warframe interface, but because necramechs are missing some fields that warframes have (like auraPolarity), it didn't make a ton of logical sense to have them extend them and have those fields silently disappear. I also briefly though about making a base interact and having those both extend it, but that came with its own set of problems as well.

TobiTenno:
why not make a WikiaNecramech and flex on if it's a necramech or warframe for what fields go in? splitting to be more exact is generally better...

Lugia00-boop:
I think i have a better idea for this now, I was getting a bit tunnel visioned on just the scraper. I looked a bit further down the line and since things end up getting processed into different wikiaData types based on their category/type. 
https://github.com/WFCD/warframe-items/blob/e754da2d337eeaea93773e63cb25e94422567363/build/parser.ts#L887-L894
While it is a slightly bigger of a change than I was initially thinking, I added a new type in the itemTypes.json for necramechs, since i think it would be nicest way to do this. I changed it so now it passes the whole wikiaNecramech[] to the parser and it uses the new type to set the "necramech" config in resolveWikiaConfig to set the type to look for in the wikiaTypes and the handler to use.
this also will allow for changing way a necramech is handled later on by just adding necramech handler if DE ever decides necramechs need some love. at the moment though the fields all seem ok, and this gets to the point of adding the wikia data in.

TobiTenno:
looks like that doesn't even pass type checks, so i'm gonna have to say you need to change stuff before i dive deeper

Lugia00-boop:
did not realize I needed to manually run the type checking before I pushed, forgot to add the new type to be allowed in the index.

TobiTenno:
yeah, since it's a post-build thing, i didn't want to enforce a fresh build on push, since that could hose your connection, especially if you just ran it, whereas tests are pretty minimal

build/wikia/transformers/transformWeapon.ts
```diff
@@ -196,6 +196,20 @@ const parseSlam = ({
   };
 };
 
+/**
+ * Convert some odd weapon naming conventions
+ * @param ToConvert - weapon to fix
+ * @return weapon with naming fixed
+*/
+const fixNamingConvention = (NameToConvert: string): string => {
+  if (NameToConvert === 'Dark Split-Sword (Heavy Blade)') {
```
TobiTenno: i _really_ don't like this since it's not even extensible without code changes. this should be a config instead of a direct if/return

i think this is my last major gripe, because having a config file makes this really easy to discover wiki name overrides as opposed to being buried in the transformer code

test/index.spec.mjs
TobiTenno: ok, i wasn't reading close enough, these follow the patterns of checking that there is no difference in deduped. for some reason i thought this was checking a different list. my mistake.

test/index.spec.mjs
TobiTenno: ok, i wasn't reading close enough, these follow the patterns of checking that there is no difference in deduped. for some reason i thought this was checking a different list. my mistake.

build/wikia/transformers/transformWeapon.ts
Lugia00-boop: I can agree with that, it is several levels deep in the code. I'll turn this into a config

TobiTenno (approved):

